PR #2, opened 2015-07-29
Don't read recursive when getting profile names.

spikecurtis:
Noticed this while working on IPAM.  We don't need to recurse into the rules and tags for profiles to get their names.


tomdee:
LGTM

PR #6, opened 2015-08-05
IPAM based on blocks and handles.

spikecurtis:


calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+    prefix = PREFIX_MASK[address.version] & address
+    prefixlen = BITS_BY_VERSION[address.version] - BLOCK_SIZE_BITS
+    block_id = "%s/%s" %(prefix, prefixlen)
+    return IPNetwork(block_id)
+
+
+class NoHostAffinityWarning(Exception):
+    """
+    Tried to auto-assign in a block this host didn't own.  This exection can be
```
robbrockbank: Typo: exection

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        Both to minimize collisions, where multiple hosts attempt to change a
+        single block, and to support route aggregation, each block has affinity
+        to a single Calico host.  That host does not hold exclusive rights to
+        modify the block; any host may still do that.  The host with affinity
+        simply uses the block as the place where it first searches if the user
+        asked to have the IP assigned automatically.
+        """
+
+        self.allocations = [None] * BLOCK_SIZE
```
robbrockbank: Better to use 0?  It'll result in a slightly smaller JSON blob.

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        A fixed length array with one entry for every address in the block.
+        None means unallocated.  A non-negative integer indicates the address
```
robbrockbank, line 83: Does a -ve integer mean anything?

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        is allocated, and is the index into the `attributes` array for the
+        attributes assigned to the allocation.
+        """
+
+        self.attributes = []
```
robbrockbank, line 88: Are the atrributes essentially arbitrary, or should be doc'ing them here?

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+
+        json_dict = {AllocationBlock.CIDR: str(self.cidr),
+                     AllocationBlock.AFFINITY:
+                         AllocationBlock.HOST_AFFINITY_T % self.host_affinity,
+                     AllocationBlock.ALLOCATIONS: self.allocations,
+                     AllocationBlock.ATTRIBUTES: self.attributes}
+        return json.dumps(json_dict)
+
+    @classmethod
+    def from_etcd_result(cls, etcd_result):
```
robbrockbank, line 112: from_json would be nicer I'd have thought... it mirrors the the to_json method, and doesn't tie this to an etcd backend.

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        A fixed length array with one entry for every address in the block.
+        None means unallocated.  A non-negative integer indicates the address
+        is allocated, and is the index into the `attributes` array for the
+        attributes assigned to the allocation.
+        """
+
+        self.attributes = []
+        """
+        Dictionaries of attributes for allocations.
```
robbrockbank: It's not a dictionary.

calico_containers/pycalico/block.py
spikecurtis, line 83: No.  They're array indexes.

calico_containers/pycalico/block.py
robbrockbank, line 112: Ah I see - you really are keeping the etcd value to use for the update.

calico_containers/pycalico/block.py
robbrockbank: Do we tidy up obsolete attributes?

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        # Process & check attributes
+        attributes = json_dict[AllocationBlock.ATTRIBUTES]
+        block.attributes = attributes
+        assert (block._verify_attributes())
+
+        return block
+
+    def update_result(self):
+        """
+        Return the EtcdResult with any changes to the object written to
```
robbrockbank, line 141: Its a minor niggle (to me at any rate) that this class is etcd specific.  Would it be worth just storing some arbitrary DB data that is used by the etcd specific client?

Prolly not worth it though?

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        # reference counting.  If we're deleting all the references, then it
+        # needs to be cleaned up.
+        attr_index_to_delete = set()
+        ref_counts = self._get_attribute_ref_counts()
+        for idx, refs in deleting_ref_counts.iteritems():
+            if ref_counts[idx] == refs:
+                attr_index_to_delete.add(idx)
+
+        # Delete attributes if necessary
+        if attr_index_to_delete:
```
robbrockbank: minor niggle:  attr_index sounds like this is just a single index.

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+
+        # All attributes updated.  Finally, release all the requested
+        # addressses.
+        for ordinal in ordinals:
+            self.allocations[ordinal] = None
+
+        return unallocated
+
+    def _delete_attributes(self, attr_index_to_delete, ordinals):
+        """
```
robbrockbank, line 374: Worth saying that this rebalances (not really the correct word) the attribute list and indexes.

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        Check if the key and attributes match existing and return the index, or
+        if they don't exist, add them and return the index.
+        """
+        assert json.dumps(attributes), \
+            "Attributes aren't JSON serializable."
+        attr = {AllocationBlock.ATTR_PRIMARY: primary_key,
+                AllocationBlock.ATTR_SECONDARY: attributes}
+        attr_index = None
+        for index, exist_attr in enumerate(self.attributes):
```
robbrockbank: should this just check the primary key and assert the attributes are the same?

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        return True
+
+
+def get_block_cidr_for_address(address):
+    """
+    Get the block ID to which a given address belongs.
+    :param address: IPAddress
+    """
+    prefix = PREFIX_MASK[address.version] & address
+    prefixlen = BITS_BY_VERSION[address.version] - BLOCK_SIZE_BITS
```
robbrockbank: BLOCK_PREFIX_LEN has been defined.

calico_containers/pycalico/block.py
spikecurtis, line 141: Yeah, it's hard to write some abstraction layer that is both simple and very likely to work for some unknown replacement to `etcd`.  I think this code is at least simple, which should lend it to being easily unpicked should the day come that we support several backends.

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+        return ref_counts
+
+    def _find_or_add_attrs(self, primary_key, attributes):
+        """
+        Check if the key and attributes match existing and return the index, or
+        if they don't exist, add them and return the index.
+        """
+        assert json.dumps(attributes), \
+            "Attributes aren't JSON serializable."
+        attr = {AllocationBlock.ATTR_PRIMARY: primary_key,
```
robbrockbank: After our chat I find this naming somewhat confusing.  Primary key just sounds like it should be a unique id of the attribute set to me.

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,383 @@
...
+    ATTRIBUTES = "attributes"
+    ATTR_PRIMARY = "primary"
+    ATTR_SECONDARY = "secondary"
+
+    def __init__(self, cidr_prefix, host_affinity):
+        assert isinstance(cidr_prefix, IPNetwork)
+        assert cidr_prefix.cidr == cidr_prefix
+
+        # Make sure the block is the right size.
+        assert cidr_prefix.prefixlen == (BITS_BY_VERSION[cidr_prefix.version] -
```
robbrockbank: BLOCK_PREFIXLEN

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+    def _compare_and_swap_block(self, block):
         """
-        Attempt to allocate an IP address from the provided pool.
+        Write the block using an atomic Compare-and-swap.
+        """
+
+        # If the block has a db_result, CAS against that.
+        if block.db_result is not None:
+            try:
+                self.etcd_client.update(block.update_result())
```
robbrockbank, line 78: If the update_result processing was actually in here there then the AllocationBlock would be non etcd specific no?

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+            except EtcdAlreadyExist:
+                raise CASError(str(block.cidr))
+        else:
+            # Block is new.  Write it with prevExist=False
+            key = _datastore_key(block.cidr)
+            value = block.to_json()
+            try:
+                self.etcd_client.write(key, value, prevExist=False)
```
robbrockbank, line 86: Do we need to update db_result here - or does the API no allow us to re-use an AllocationBlock object without first re-querying it?

robbrockbank:
Need to update the etcd document.

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+        # If pool specified, filter to only include ones in the pool.
+        if pool is not None:
+            assert isinstance(pool, IPPool)
+            block_ids = [cidr for cidr in block_ids if cidr in pool]
+
+        return block_ids
+
+    def _new_affine_block(self, host, version, pool):
+        """
+        Create an register a new affine block for the host.
```
robbrockbank: an->and

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+    :return: etcd key as string.
+    """
+    path = IPAM_BLOCK_PATH % {'version': block_cidr.version}
+    return path + str(block_cidr).replace("/", "-")
+
+
+class IPAMClient(BlockReaderWriter):
+
+    def auto_assign_ips(self, num_v4, num_v6, primary_key, attributes,
+                        pool=(None, None)):
```
robbrockbank: num_v4 and num_v6 are separate parms.  Any reason why pool is a tuple for v4 and v6?

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+            num_remaining = num - len(allocated_ips)
+
+        # If there are still addresses to allocate, then we've run out of
+        # blocks with affinity.  Try to fullfil address request by allocating
+        # new blocks.
+        retries = RETRIES
+        while num_remaining > 0 and retries > 0:
+            retries -= 1
+            try:
+                new_block = self._new_affine_block(my_hostname,
```
robbrockbank: Should retries be reset after successfully creating a new affine block?

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+        # with some affinity to us, and tried (and failed) to allocate new
+        # ones.  Our last option is a random hunt through any blocks we haven't
+        # yet tried.
+        # TODO: SJC add this function.
+
+        return allocated_ips
+
+    def _auto_assign_block(self, block_id, num, primary_key, attributes):
+        """
+        Automatically pic IPs from a block and commit them to the data store.
```
robbrockbank: pic->pick

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+                # No more blocks.
+                break
+            ips = self._auto_assign_block(new_block,
+                                          num_remaining,
+                                          primary_key,
+                                          attributes)
+            allocated_ips.extend(ips)
+            num_remaining = num - len(allocated_ips)
+        if retries == 0:  # pragma: no cover
+            raise RuntimeError("Hit Max Retries.")
```
robbrockbank, line 488: Runtime, or better to have a OySomeoneIsClashingWithMyRequestsError?

robbrockbank, line 488: strictly if retries ==0 and num_remaining > 0   ?

Should we be attempting tidy up if this happens .. or is that why you are doing a RuntimeException because basically all bets are off?

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+                           block_cidr)
+                return
+
+            # Some other host beat us to claiming this block.  Clean up.
+            self.etcd_client.delete(key)
+
+            # Throw a key error to let the caller know the block wasn't free
+            # after all.
+
+            raise KeyError(block_id)
```
robbrockbank: KeyError doesn't feel correct for this.

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+                pools = self.get_ip_pools(address.version)
+                if any([address in pool for pool in pools]):
+                    # Address is in a pool.  Create and claim the block.
+                    try:
+                        self._claim_block_affinity(my_hostname,
+                                                   block_cidr)
+                    except KeyError:
+                        # Happens if something else claims the block between
+                        # the read above and claiming it.
+                        continue
```
robbrockbank: Why continue?  Are we hoping that someone will release it again?

calico_containers/pycalico/ipam.py
```diff
@@ -15,78 +15,472 @@
...
+                # Block doesn't exist.  Is it in a valid pool?
+                pools = self.get_ip_pools(address.version)
+                if any([address in pool for pool in pools]):
+                    # Address is in a pool.  Create and claim the block.
+                    try:
+                        self._claim_block_affinity(my_hostname,
+                                                   block_cidr)
+                    except KeyError:
+                        # Happens if something else claims the block between
+                        # the read above and claiming it.
```
robbrockbank: log here

djosborne:
This PR doesn't take into account special IP addresses (such as the broadcast address) which probably should not be assigned to containers when `auto_assign` is called

calico_containers/pycalico/ipam.py
spikecurtis, line 488: Yeah, all bets are off in this case.  The number of retries is meant to be so high that we would never hit it even under heavy load.  If multiple threads are trying to allocate from the same block one of them always wins the race, so the algorithm will make forward progress.

This error is meant to prevent code bugs from looping forever.

calico_containers/pycalico/ipam.py
spikecurtis: Fixed.

spikecurtis:
@tomdee Rob has reviewed the basic code and written UTs for the handle code.  We should chat about what further review is needed before we're comfortable merging.  Kubernetes and Mesos plug-ins are using the new code.

calico_containers/pycalico/block.py
```diff
@@ -0,0 +1,494 @@
...
+            # Increment our count of addresses by handle.
+            handle_id = self.\
+                attributes[attr_idx][AllocationBlock.ATTR_HANDLE_ID]
+            handle_count = handles_with_counts.setdefault(handle_id, 0)
+            handle_count += 1
+            handles_with_counts[handle_id] = handle_count
+
+        # Compute which attributes need to be cleaned up.  We do this by
+        # reference counting.  If we're deleting all the references, then it
+        # needs to be cleaned up.
```
Symmetric, line 278: Can you add a quick comment to describe whether race conditions in the cleanup step are an issue here please? Did a bit of digging but it's not immediately clear how you handle the case where another process added a new reference after this process fetched its state from etcd. I think that's probably worth documenting.

Symmetric:
A general comment -- what about logging? Are we still not making any effort to generate logs in our core components? I think we should be doing so on any new development (even if we're not retroactively backfilling missing logs).

Symmetric:
Apart from those two comments (up to you whether you want to address them), LGTM

calico_containers/pycalico/block.py
spikecurtis, line 278: Added note to class-level comment

PR #8, opened 2015-08-07
Add methods used by container_add/container_remove

luke-mino-altherr:


calico_containers/pycalico/datastore.py
```diff
@@ -745,6 +745,43 @@ def update_endpoint(self, endpoint):
         endpoint._original_json = new_json
 
     @handle_errors
+    def create_endpoint(self, hostname, orchestrator_id, workload_id,
```
caseydavenport: You'll want to mention in the docstring that this does not store the endpoint in ETCD, as that should be done with endpoint_set or endpoint_update, as appropriate.

calico_containers/pycalico/ipam.py
```diff
@@ -85,12 +85,22 @@ def assign_address(self, pool, address):
...
         """
+        if pool is None:
+            pools = self.get_ip_pools(address.version)
+            for canidate_pool in pools:
+                if address in candidate_pool:
+                    pool = candidate_pool
+                    break
+
+        if pool is None:
+            raise KeyError("IPPool is not found")
```
caseydavenport: "IP address %s does not belong to any configured pools"

caseydavenport:
You'll need some UTs for these guys.

calico_containers/pycalico/netns.py
```diff
@@ -180,6 +179,28 @@ def remove_ip_from_ns_veth(container_pid, ip, veth_name_ns):
                        "dev", "%(device)s" % veth_name_ns])
 
 
+def provision_veth(self, endpoint, pid, ip, veth_name_ns):
```
ozdanborne: Can we rename `pid` to `ns_pid`? Elsewhere in pycalico we refer to `container_pid`, `pid`, and `cpid` and the ambiguity is starting to get a little crazy. 

Limit this change only to the new functions you've introduced - no need to go running around the rest of the code base changing everything else right now

djosborne: Does Kubernetes support setting multiple IP addresses to a veth?

calico_containers/pycalico/netns.py
```diff
@@ -180,6 +179,28 @@ def remove_ip_from_ns_veth(container_pid, ip, veth_name_ns):
...
+
+    :param endpoint: The endpoint object to provision the veth on
+    :param pid: The PID of the namespace to operate in
+    :param ip: The IPAddress to provision.
+    :param veth_name_ns: The name of the interface in the namespace
+    :return The mac address of the veth as a string
+    """
+    create_veth(endpoint.name, endpoint.temp_interface_name)
+    move_veth_into_ns(pid, endpoint.temp_interface_name, veth_name_ns)
+    add_ip_to_ns_veth(pid, ip, veth_name_ns)
```
ozdanborne: We should be getting the IPs that have been saved into the endpoint, and applying those to the veth. Change this to:

```
for ip in ep.ipv4_nets:
    add_ip_to_ns_veth(pid, ip, veth_name_ns)
```

caseydavenport: And ipv6 as well?

Also note that this change affects the lines below it where the default routes are set.  Ultimately, we should not need to pass in the ip argument to this method.

calico_containers/pycalico/ipam.py
```diff
@@ -112,12 +122,22 @@ def unassign_address(self, pool, address):
...
 
         :param IPPool or IPNetwork pool: The pool that the assignment is from.
+        If the pool is None, get the pool from datastore
         :param IPAddress address: The address to unassign.
-
         :return: True if the address was unassigned, false otherwise. An
         exception is thrown for any error conditions.
         :rtype: bool
         """
+        if pool is None:
```
ozdanborne, line 128: This functionality is duplicated once. Does it make sense to throw it in a function, and then 

```
pool = pool or get_pool(address)
```

Sort of nitpicky, might just be personal preference there...

calico_containers/pycalico/ipam.py
```diff
@@ -85,12 +85,22 @@ def assign_address(self, pool, address):
...
 
         :param IPPool or IPNetwork pool: The pool that the assignment is from.
+        If pool is None, get the pool from datastore
         :param IPAddress address: The address to assign.
-
         :return: True if the allocation succeeds, false otherwise. An
         exception is thrown for any error conditions.
         :rtype: bool
         """
+        if pool is None:
```
caseydavenport, line 96: This looks like a good candidate to break out into a helper.

Something like

```
def get_pool(self, ip):
  """
  Returns the pool which contains the given IP address.
  """
  ....
```

calico_containers/pycalico/ipam.py
```diff
@@ -112,12 +122,22 @@ def unassign_address(self, pool, address):
...
+            self.get_ip_pools(address.version)
+            for canidate_pool in pools:
+                if address in candidate_pool:
+                    pool = candidate_pool
+                    break
+
+        if pool is None:
+            raise KeyError("IPPool is not found")
```
ozdanborne: "IP Address does not belong to any configured Pools"

ozdanborne:
That's all I got for now!

calico_containers/pycalico/netns.py
```diff
@@ -180,6 +179,32 @@ def remove_ip_from_ns_veth(container_pid, ip, veth_name_ns):
                        "dev", "%(device)s" % veth_name_ns])
 
 
+def provision_veth(self, endpoint, ns_pid, veth_name_ns):
```
Symmetric: For rkt, we'll need this method to take `ns_path` instead of `ns_pid`. That will mean rewiring a few functions in the netns.py module (including the NamedNamespace class). I suggest creating a get_docker_ns_path(pid) helper function to encapsulate the pid=>path logic for the docker case. We get the full path in the rkt case, so we don't need a helper method there.

Symmetric: Note, that change is somewhat unrelated to this PR, so I'm happy to make that change myself if you want to merge this as is.

calico_containers/pycalico/ipam.py
```diff
@@ -85,12 +85,17 @@ def assign_address(self, pool, address):
...
+        If pool is None, get the pool from datastore
         :param IPAddress address: The address to assign.
-
         :return: True if the allocation succeeds, false otherwise. An
         exception is thrown for any error conditions.
         :rtype: bool
         """
+        pool = pool or self.get_pool(address)
+        if pool is None:
+            raise KeyError("IP address %s does not belong to any configured"
```
caseydavenport: Instead of `KeyError`, we might want to define an API error for these sorts of things.

`InvalidPoolError` or something like that?  @robbrockbank any thoughts?

calico_containers/pycalico/ipam.py
```diff
@@ -112,12 +117,17 @@ def unassign_address(self, pool, address):
...
 
         :param IPPool or IPNetwork pool: The pool that the assignment is from.
+        If the pool is None, get the pool from datastore
         :param IPAddress address: The address to unassign.
-
         :return: True if the address was unassigned, false otherwise. An
         exception is thrown for any error conditions.
         :rtype: bool
         """
+        pool = pool or self.get_pool(address)
```
robbrockbank, line 127: Do we need to worry about the assignment being in a different pool since get_pool only returns a single pool and there could potentially be overlapping pools I think (I don't think we police against that at the moment).

calico_containers/pycalico/ipam.py
robbrockbank: Yeah might be nice.

Incidently, Spike is working on an IPAM re-work -- might be worth checking that this doens't conflict.

calico_containers/pycalico/ipam.py
ozdanborne, line 127: I don't think calico knows how to handle overlapping pools anywhere. Might be worth raising a backlog issue to define which pool should be chosen (probably default to the most specific subnet)

calico_containers/pycalico/ipam.py
Symmetric: All of these exceptions should descend from LibcalicoError or some other base class -- so that we can catch all API errors if we don't care exactly what happened.

calico_containers/pycalico/netns.py
```diff
@@ -180,6 +179,32 @@ def remove_ip_from_ns_veth(container_pid, ip, veth_name_ns):
...
+    should update the endpoint mac with the mac address provided
+    by the function and then call update_endpoint
+
+    :param endpoint: The endpoint object to provision the veth on
+    :param ns_pid: The PID of the namespace to operate in
+    :param veth_name_ns: The name of the interface in the namespace
+    :return The mac address of the veth as a string
+    """
+    create_veth(endpoint.name, endpoint.temp_interface_name)
+    for ip in endpoint.ipv4_nets | endpoint.ipv6_nets:
```
Symmetric: ipvX_nets does not contain ips, it contains networks (i.e. subnets).

ip should be an argument to this method; it's not stored on the endpoint..

calico_containers/pycalico/ipam.py
spikecurtis, line 127: I'm reworking all this stuff so that we won't have overlapping pools.

calico_containers/pycalico/netns.py
ozdanborne: Ooh that's a significant design flaw with this whole approach. For mesos, the API call which creates the interface is passed the desired IP, but the API call which moves the interface into the NS does not - so it is "necessary" in that situation to pull the IP from the endpoint, as the second API call that calls `provision_veth` will not know what IP address to pass it.

@Symmetric Is there a reason that the libnetwork design approach doesn't bind an IP to an endpoint? Can we?

spikecurtis: @Symmetric is incorrect.  The networks are _not_ subnets.  They will be /32 networks (for IPv4) from which you can easily get the IP.

spikecurtis: Putting a little more meat on the bone, ipv_X__nets are, for now, just the IP addresses, but encoded as networks for future-proofing the data model.

In particular, IPv6 SLAAC assigns a /64 to each host, allowing for multiple interface IDs without changing the routing behavior.

calico_containers/pycalico/netns.py
```diff
@@ -180,6 +179,32 @@ def remove_ip_from_ns_veth(container_pid, ip, veth_name_ns):
...
+    by the function and then call update_endpoint
+
+    :param endpoint: The endpoint object to provision the veth on
+    :param ns_pid: The PID of the namespace to operate in
+    :param veth_name_ns: The name of the interface in the namespace
+    :return The mac address of the veth as a string
+    """
+    create_veth(endpoint.name, endpoint.temp_interface_name)
+    for ip_net in endpoint.ipv4_nets | endpoint.ipv6_nets:
+        move_veth_into_ns(ns_pid, endpoint.temp_interface_name, veth_name_ns)
```
tomdee: You're moving the veth into the ns multiple times. That can't be right...

calico_containers/pycalico/datastore.py
```diff
@@ -258,6 +259,24 @@ def get_ip_pools(self, version):
         return pools
 
     @handle_errors
+    def get_pool(self, ip):
+        """
+        Returns the pool which contains the given IP address
```
tomdee: Returns the _first_ pool that contains the given IP address.

calico_containers/pycalico/datastore.py
```diff
@@ -745,6 +764,46 @@ def update_endpoint(self, endpoint):
         endpoint._original_json = new_json
 
     @handle_errors
+    def create_endpoint(self, hostname, orchestrator_id, workload_id,
```
tomdee, line 767: Shouldn't this take a list of IP addresses?

calico_containers/pycalico/netns.py
luke-mino-altherr: @Symmetric Do you agree with Spike? I fixed the way I am accessing the IP addresses (indicating .ip on the IPNetwork obj), however, I am still pulling the IPs from the endpoint's ipvX_nets.

Symmetric: Yep. You should probably add a comment in create_endpoint explaining why
you're creating a network since it's non-obvious.

Bonus points for adding a comment to datastore.py (perhaps with Spike's
explanation); I can't find an explanation for the translation to a network
anywhere in the code.

On Mon, Aug 10, 2015 at 4:47 PM, Luke Mino-Altherr <notifications@github.com

calico_containers/pycalico/datastore.py
```diff
@@ -745,6 +764,46 @@ def update_endpoint(self, endpoint):
...
+                      mac=mac)
+        next_hops = self.get_default_next_hops(hostname)
+
+        for ip in ip_list:
+            try:
+                next_hop = next_hops[ip.version]
+            except KeyError:
+                raise AddrFormatError("This node is not configured for IPv%d."
+                                       % ip.version)
+            network = IPNetwork(IPAddress(ip))
```
robbrockbank: No need to wrap IP in IPAddress().  You're already accessing ip.version above, so ip should already be an IPAddress object.

Maybe worth asserting this is the case.

calico_containers/pycalico/datastore_errors.py
```diff
@@ -51,4 +51,10 @@ class MultipleEndpointsMatch(Exception):
     """
     More than one endpoint was found for the specified criteria.
     """
+    pass
+
+class InvalidPoolError(Exception):
```
robbrockbank: Can you rename to something more specific - this makes it sound like the pool was invalid, but that's not the case - you only throw this exception when there is no matching Pool.

You could also pass in the IP address as a parameter, so that whoever catches it knows what IP address did not match any pool  (but that might not be necessary if the context is such that we can always tell which IP address it is).

calico_containers/pycalico/netns.py
```diff
@@ -180,6 +179,32 @@ def remove_ip_from_ns_veth(container_pid, ip, veth_name_ns):
                        "dev", "%(device)s" % veth_name_ns])
 
 
+def provision_veth(endpoint, ns_pid, veth_name_ns):
+    """
+    Create the veth, move into the container namespace, add the IP and
+    set up the default routes.
+
+    Note, the endpoint will not be updated in ETCD. If desired, the user
```
robbrockbank: etcd is lowercase

calico_containers/pycalico/netns.py
```diff
@@ -180,6 +179,32 @@ def remove_ip_from_ns_veth(container_pid, ip, veth_name_ns):
                        "dev", "%(device)s" % veth_name_ns])
 
 
+def provision_veth(endpoint, ns_pid, veth_name_ns):
```
robbrockbank: We'd previously had a tidy up of the netns class to remove details about endpoints - the idea being that netns provided generic useful helper functions without really needing to understand the various data structures or anything to do with etcd.

I think this might be worth moving in to the Endpoint object, so you can just call:

endpoint.provision_veth()

robbrockbank:
Looks like you have merge conflicts that will need to be resolved.

robbrockbank:
lgtm

PR #9, opened 2015-08-12
Allow specifying rules when creating profile.

Symmetric:
Rather than having to create a profile with the default rules,
then change those rules, allow the profile's starting rules
to be specified at creation time.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -425,7 +425,7 @@ class DatastoreClient(object):
             return True
 
     @handle_errors
-    def create_profile(self, name):
+    def create_profile(self, name, rules=None):
         """
         Create a policy profile.  By default, endpoints in a profile
         accept traffic only from other endpoints in that profile, but can send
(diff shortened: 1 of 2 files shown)


calico_containers/pycalico/datastore.py
```diff
@@ -434,6 +434,8 @@ def create_profile(self, name):
         Note this will clobber any existing profile with this name.
 
         :param name: Unique string name for the profile.
+        :param rules: The rules to set on the profile
+        :type rules Rules
```
spikecurtis, line 439: This docstring needs to explain that `rules` is optional and what happens if it isn't specified.

tomdee:
@Symmetric Are you going to fix/merge this?

Symmetric:
Planning to, not at the top of my list (currently sprinting to get some CNI
stuff in). Revisit next week?

Symmetric:
Updated PR with a comment to address Spike's concern, and force-pushed to keep history clean.

PTAL

tomdee:
LGTM - merging

PR #11, opened 2015-08-18
Use pid and path

tomdee:


robbrockbank:
LGTM.

As discussed, aim is to get a more architectural change out during the next week.

PR #13, opened 2015-08-18
Abstract Namespace specification to allow path/pid

caseydavenport:
For rkt, we need to be able to specify namespaces whose file
is at an arbitrary path, in addition to those at
/proc/<pid>/ns/net.

This commit adds an abstraction layer to allow specifying
namespaces at arbitrary locations.

### DIFF
--- a/calico_containers/pycalico/netns.py
+++ b/calico_containers/pycalico/netns.py
@@ -104,17 +104,18 @@ def veth_exists(veth_name_host):
             return False
 
 
-def move_veth_into_ns(cpid, veth_name_ns_temp, veth_name_ns):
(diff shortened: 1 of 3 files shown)


calico_containers/pycalico/netns.py
```diff
@@ -283,3 +294,17 @@ class NamespaceError(Exception):
...
     pass
+
+
+class Namespace(object):
+    def __init__(self, path):
+        self.path_ = path
+
+    @property
+    def path(self):
+        return self.path_
```
TrimBiggs: I don't think this gives us any benefits.  The init should just be `self.path = path`.

TrimBiggs:
One minor comment then LGTM

PR #18, opened 2015-08-25
Only enabled IPIP when Pool added that uses it

robbrockbank:


calico_containers/pycalico/datastore.py
```diff
@@ -82,8 +82,12 @@
 DEFAULT_LOG_SEVERITY_SCREEN = "info"
 DEFAULT_LOG_FILE_PATH = "none"
 
-# Default global IP in IP.
-DEFAULT_IP_IN_IP = "true"
+# Default global IP in IP.  By default this is turned off and turned on as
```
TrimBiggs: Wording is kind of awkward here, sounds like it is 'turned off and turned on'.
Maybe saying "This is off by default and is only turned on when required by an IP Pool."

calico_containers/pycalico/datastore.py
```diff
@@ -336,6 +340,14 @@ def add_ip_pool(self, version, pool):
         assert version in (4, 6)
         assert isinstance(pool, IPPool)
 
+        # If IP in IP is enabled, ensure that it is enabled globally, and if
+        # not enable it.
```
TrimBiggs: The "and if not enable it" sounds funny, I think it can just be removed since ensuring something is enabled implies you will enable it if necessary.

calico_containers/pycalico/datastore.py
```diff
@@ -82,8 +82,12 @@
...
 DEFAULT_LOG_FILE_PATH = "none"
 
-# Default global IP in IP.
-DEFAULT_IP_IN_IP = "true"
+# Default global IP in IP.  By default this is turned off and turned on as
+# soon as it is required by an IP Pool.
+DEFAULT_IP_IN_IP = "false"
+
+# IP in IP enabled value.
+IP_IN_IP_ENABLED = "true"
```
TrimBiggs, line 88: Maybe have `IP_IN_IP_DISABLED` also and set `DEFAULT_IP_IN_IP = IP_IN_IP_DISABLED`

TrimBiggs:
A couple of minor comments, otherwise LGTM

PR #20, opened 2015-08-28
Correct duplicate test name

alexwlchan:
There were two tests in this file named `test_get_host_ips_exclude_docker`, which meant that the first test would never run.

The quick-and-easy fix is to rename one of the tests, which is what I've done. Somebody who knows the tests better than me should probably check:
1. Whether we actually want both tests (they're similar but not identical)
2. If so, whether I've chosen an appropriate name


tomdee:
Looks fine.

PR #24, opened 2015-09-25
Raise EtcdValueErrors on CAS errors

alexhersh:
Narrow Exceptions to etcd specific handlers.

### DIFF
--- a/calico_containers/pycalico/ipam.py
+++ b/calico_containers/pycalico/ipam.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from etcd import EtcdKeyNotFound, EtcdAlreadyExist
+from etcd import EtcdKeyNotFound, EtcdAlreadyExist, EtcdCompareFailed
 
 from netaddr import IPAddress, IPNetwork
 import socket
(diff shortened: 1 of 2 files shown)


calico_containers/pycalico/ipam.py
```diff
@@ -76,7 +76,7 @@ def _compare_and_swap_block(self, block):
             _log.debug("CAS Update block %s", block)
             try:
                 self.etcd_client.update(block.update_result())
-            except ValueError:
+            except EtcdValueError:
```
caseydavenport: To be as specific as possible, shouldn't these all be catching `EtcdCompareFailed`?

alexhersh: I was worried that ValueError would catch more than the Compare Error that I expected, but looking again at the Exception hierarchy, I think you are right, we can afford to be more specific.

calico_containers/pycalico/ipam.py
```diff
@@ -322,7 +322,7 @@ def _compare_and_swap_handle(self, handle):
                 _log.debug("Handle %s is not empty.", handle.handle_id)
                 try:
                     self.etcd_client.update(handle.update_result())
-                except ValueError:
+                except EtcdCompareFailed:
                     raise CASError(handle.handle_id)
         else:
             _log.debug("CAS Write new handle %s", handle.handle_id)
```
caseydavenport, line 328: Will write(...) throw an EtcdCompareFailed?  Do we need to catch it below?

alexhersh, line 328: No, I think they only trigger when the block has a db_result to compare against (update not write).

caseydavenport:
LGTM

PR #25, opened 2015-09-25
IPAM allocation blocks of 64 addresses

spikecurtis:


calico_containers/pycalico/ipam.py
```diff
@@ -33,7 +33,7 @@
 
 _log = logging.getLogger(__name__)
 
-IPAM_V_PATH = "/calico/ipam/v1/"
+IPAM_V_PATH = "/calico/ipam/v2/"
```
caseydavenport, line 36: Have we estimated how much work it would be to make the IPAM code resilient to different block sizes?

If it isn't an excessive amount, it might save us down the road to support existing blocks, and any new blocks will be 64 addresses in size.

spikecurtis, line 36: It's a lot of work.  Off the cuff, I'd guess at least a developer-month by the time we design, implement, UT and ST.

caseydavenport, line 36: Yikes, that is more than I expected.  Probably not worth doing right now then.

caseydavenport:
LGTM

PR #27, opened 2015-09-29
Allow `add_ip_pool` to write the IP_IN_IP etcd key when `/calico` empty

alexhersh:
- add_ip_pool currently performs a read before it decides to write to the global IP_IN_IP key
- In ipip deployments, we recommend that users call `pool add` before `node`, meaning that `pool add` needs to be able to write to an empty etcd

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -17,7 +17,7 @@ import os
 import socket
 import uuid
 import etcd
-from etcd import EtcdKeyNotFound, EtcdException, EtcdNotFile
+from etcd import EtcdKeyNotFound, EtcdException, EtcdNotFile, EtcdKeyError
 
 from netaddr import IPNetwork, IPAddress, AddrFormatError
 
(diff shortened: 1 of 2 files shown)


calico_containers/pycalico/datastore.py
```diff
@@ -341,8 +341,11 @@ def add_ip_pool(self, version, pool):
         # If IP in IP is enabled on the pool, ensure that it is enabled
         # globally.
         if pool.ipip:
-            result = self.etcd_client.read(IP_IN_IP_PATH)
-            if result.value != IP_IN_IP_ENABLED:
+            try:
+                result = self.etcd_client.read(IP_IN_IP_PATH)
+            except EtcdException:
```
caseydavenport: Is this the correct Exception to be catching here?  There are a couple of failure cases here at least.

1) The key does not exist yet in `etcd` (this is the one we care about)
2) We fail to connect to `etcd` at all (we don't want to catch this)

calico_containers/pycalico/datastore.py
```diff
@@ -341,8 +341,11 @@ def add_ip_pool(self, version, pool):
         # If IP in IP is enabled on the pool, ensure that it is enabled
         # globally.
         if pool.ipip:
-            result = self.etcd_client.read(IP_IN_IP_PATH)
-            if result.value != IP_IN_IP_ENABLED:
+            try:
```
caseydavenport, line 362: Could you add a comment describing the purpose of this try/except, and why we perform a read first at all?

caseydavenport:
I suspect this needs a UT.

spikecurtis:
LGTM.  Can you squash commits?

PR #28, opened 2015-09-30
Add workload_exists method

### LINKED ISSUE #501: Can't remove a container from Calico if it's already been killed and rmed from Docker
I wanted to stop a Calico node (`calicoctl node stop`), but it complained (rightly) that there was still a workload running.  I `docker kill`ed and `docker rm`ed the container, but it remained in calicoctl.

I then tried `calicoctl container remove <name>`, `calicoctl container remove <short id from docker PS>` and `calicoctl container remove <long workload id from calicoctl endpoint show --detailed>`, but in all cases it returned "Container <identifier> was not found.".

I was using calico-docker 0.6.0, having followed the Ubuntu manual install process.  I was able to `calicoctl node stop --force`, but it should be possible to remove a container from Calico without doing this.

tomdee:
Required for https://github.com/projectcalico/calico-docker/pull/507 (which fixes https://github.com/projectcalico/calico-docker/issues/501)

When using calicoctl to remove workloads, sometimes it might find that docker isn't running but we should still let the user remove the workload from etcd. 

This function supports that by having logic which say "the user has passed in an exact workload_id which exists in etcd, there is no need to check with docker"


spikecurtis:
@tomdee Can you describe what this is used for & why we need it?  I ask because we're expanding the `libcalico` API with this PR.

tomdee:
@spikecurtis see above. Would you mind reviewing?

tomdee:
No longer need this for the change I'm making

PR #30, opened 2015-10-06
Set up host affinity path making code match comment.

plwhite:
Can I get some feedback on this?

Background is that there was a comment saying that libcalico would set up the IPAM directories, but it didn't - and this broke my route aggregation code in calico-docker. I put in a trivial fix, but there may be style issues.

I've not written any new tests because I haven't been able to run them, but if you can point me at some instructions I can extend the tests. However, seems worth getting a sanity check now before I expend time on the tests.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -14,9 +14,10 @@
 
 import json
 import os
+import socket
 import uuid
 import etcd
-from etcd import EtcdKeyNotFound, EtcdException
+from etcd import EtcdKeyNotFound, EtcdException, EtcdNotFile
 
 from netaddr import IPNetwork, IPAddress, AddrFormatError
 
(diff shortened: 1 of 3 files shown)


calico_containers/pycalico/datastore.py
```diff
@@ -129,7 +137,15 @@ def ensure_global_config(self):
 
         # Configure IPAM directory structures (to ensure confd is able to
         # watch appropriate directory trees).
-        self.etcd_client.write(CONFIG_IF_PREF_PATH, IF_PREFIX)
+        host = socket.gethostname()
```
spikecurtis, line 140: Comment was wrong, but I'm guessing we still need the 

```
self.etcd_client.write(CONFIG_IF_PREF_PATH, IF_PREFIX)
```

What this does is configure Felix to look for the right interface prefix for container veths.

calico_containers/pycalico/#datastore.py#
```diff
@@ -0,0 +1,988 @@
+# Copyright 2015 Metaswitch Networks
```
spikecurtis: Backup cache accidentally made it into this PR.

spikecurtis:
Fix and style seem OK to me.

UTs require Docker, and then you should be able to `make ut` to get them to run in a Docker container.

plwhite:
Thanks for that. I got the UTs running, and have written some that cover this case, but need to clean them up.

Note that I stupidly managed to do my testing on the same branch which I used for this PR (doh) so the change history looks crazy. I will resolve that before I reoffer it (and will also look into your code comment - I thought I disagreed before, but will rethink).

calico_containers/pycalico/datastore.py
plwhite, line 140: I don't think we do, because of the line "write_global_config" above that writes the exact same path and value; unless I'm misunderstanding (not without precedent!) this line is effectively a duplicate.

plwhite:
No code changes since you last saw it, but I have added tests. I think the line adding the interface prefix that I removed was a duplicate, so I was right to remove it, so let me know if you disagree.

I think this is ready for merge, and assigning to @spikecurtis who looked at it last.

calico_containers/pycalico/datastore.py
spikecurtis, line 140: Nope, you're right.  I see now where the UTs check this.

spikecurtis:
LGTM :shipit:

PR #35, opened 2015-10-13
Removed use of check_call in netns and improved logging

ozdanborne:
Replaced `check_call` with `check_output`, which, when its call throws an exception, will include  a `.output` field which grabs stdout and can be used to better log and debug encountered errors


tomdee:
LGTM

PR #36, opened 2015-10-15
Update remove_host to delete IPAM blocks for host

### LINKED ISSUE #519: calicoctl node stop removes enpoint configuration but leaves IPAM data in place
This leads to dangling IP allocations

alexhersh:
- Help fix projectcalico/calico-docker#519: calicoctl node stop does not release IPAM allocation for host
- Use BGP Path format for IPAM Path constant

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -90,7 +90,8 @@ IP_IN_IP_ENABLED = "true"
 
 # IPAM paths
 IPAM_V_PATH = "/calico/ipam/v2/"
-IPAM_HOST_PATH = IPAM_V_PATH + "host/%(hostname)s/"
+IPAM_HOSTS_PATH = IPAM_V_PATH + "host/"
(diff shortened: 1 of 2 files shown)


robbrockbank:
On its own, it isn't sufficient to just delete the host IPAM tree as the IPs associated with that host will still be assigned - so this will result in inconsistent IPAM data.

I'm planning to have a think about the broader implications of node stop, so I'll revisit this soon.

robbrockbank:
Also, this code should really be in the ipam module rather than the datastore.

Probably need a delete_host or similar method in ipam to delete the IP assignments, remove any block affinities, and delete the host tree.

Don't update this PR at the moment as I'm looking at a reworked design for this.

alexhersh:
Should we close this PR if you are working on an alternate, more robust design?

robbrockbank:
Yup.  Closing now - just wanted to get my thoughts down first.

PR #38, opened 2015-10-19
Made hostname hard requirement, removed default read of socket.gethostname

ozdanborne:
Added hostname as parameter for `auto_assign`. This resolves a bug where hosts which claim affinity under a different hostname than what `socket.gethostname` returns couldn't auto_assign. 

Removed the default hostname to prevent libcalico users from encountering this bug.


ozdanborne:
After some discussion, it seems more appropriate to change libcalico to use a HOSTNAME environment variable instead of relying on orchestrators to pass the hostname variable around.

PR #40, opened 2015-10-21
Replace socket.gethostname with $HOSTNAME

### LINKED ISSUE #39: Use $HOSTNAME environment variable instead of `socket.gethostname()`
Lots of `libcalico` API calls use `socket.gethostname()` to get the name of the current host.  This is fine in most circumstances, but limits flexibility in some environments.

In particular, we actually only care that the hostname is a unique, stable identifier for the compute host.  Some orchestration systems, like Mesos, use reverse-name-resolution to name the compute host.  This can leave Calico inconsistent with the orchestrator (e.g. Mesos has FQDN and Calico just has a local name).

After discussing with @tomdee, we agreed a good solution would be to use `$HOSTNAME` instead of `socket.gethostname()`.  It's a standard environment variable that is ordinarily set to the same value as would be returned by `socket.gethostname()`.  But, it gives an admin more flexibility, since they can override an environment variable without affecting other systems and processes.

ozdanborne:
Fixes #39

### DIFF
--- a/calico_containers/pycalico/ipam.py
+++ b/calico_containers/pycalico/ipam.py
@@ -32,12 +32,12 @@ from pycalico.block import (AllocationBlock,
                             AddressNotAssignedError)
 from pycalico.handle import (AllocationHandle,
                              AddressCountTooLow)
+from pycalico.util import get_hostname
 
 _log = logging.getLogger(__name__)
 
 RETRIES = 100
 
-my_hostname = socket.gethostname()
 
 class BlockHandleReaderWriter(DatastoreClient):
     """
@@ -375,7 +375,7 @@ def _handle_datastore_key(handle_id):
 class IPAMClient(BlockHandleReaderWriter):
 
     def auto_assign_ips(self, num_v4, num_v6, handle_id, attributes,
-                        pool=(None, None), hostname=my_hostname):
+                        pool=(None, None), hostname=None):
         """
         Automatically pick and assign the given number of IPv4 and IPv6
         addresses.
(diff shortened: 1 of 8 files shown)


calico_containers/pycalico/util.py
```diff
@@ -71,3 +75,9 @@ def get_host_ips(version=4, exclude=None):
 
     return ip_addrs
 
+def get_hostname():
+    try:
+        hostname = os.environ[HOSTNAME_ENV]
+    except KeyError:
+        hostname = socket.gethostname()
```
ozdanborne: Does defaulting to `socket.gethostname` make sense from a design perspective?

alexhersh: I don't have much input on general design, but the more concise way to do this might be 

```
hostname = os.environ.get(HOSTNAME_ENV, socket.gethostname())
```

`os.environ.get` will return the default value if the key is not found

spikecurtis: I think we should fail hard if the environment variable isn't set.  Much easier to understand for the user than defaulting.  HOSTNAME is a standard environment variable, so I don't think we're putting any significant config burden on end users.

calico_containers/pycalico/ipam.py
```diff
@@ -391,14 +391,16 @@ def auto_assign_ips(self, num_v4, num_v6, handle_id, attributes,
         :param pool: (optional) tuple of (v4 pool, v6 pool); if supplied, the
         pool(s) to assign from,  If None, automatically choose a pool.
         :param hostname: (optional) the hostname to use for affinity in
-        assigning IP addresses.  Defaults to the local hostname as returned by
-        socket.gethostname().
+        assigning IP addresses.  Defaults to the $HOSTNAME environment variable.
```
tomdee: Shouldn't this just say "Defaults to the hostname returned by get_hostname()

calico_containers/pycalico/ipam.py
```diff
@@ -569,11 +571,13 @@ def assign_ip(self, address, handle_id, attributes, hostname=my_hostname):
         be JSON serializable.
         :param hostname: (optional) the hostname to use for affinity if the
         block containing the IP address has no host affinity.  Defaults to the
-        local hostname as returned by socket.gethostname().
+        $HOSTNAME environment variable.
```
tomdee: As above

calico_containers/pycalico/util.py
```diff
@@ -71,3 +75,8 @@ def get_host_ips(version=4, exclude=None):
 
     return ip_addrs
 
+def get_hostname():
+    try:
```
tomdee, line 88: Can you comment this. Both a docstring and the reason why it falls back to the socket.gethostname()

tomdee:
A few comments to address, otherwise LGTM.

PR #41, opened 2015-10-22
Added function which increments the metric of the default route

ozdanborne:
Adds functionality which can be used to Fix [#540](https://github.com/projectcalico/calico-docker/issues/540)


calico_containers/pycalico/netns.py
```diff
@@ -51,6 +51,20 @@ def setup_logging(logfile, level=logging.INFO):
...
+    """
+    Increment the metric of any default route that has a metric of 0.
+    :param namespace: The Networking namespace of the container.
+    :return: None. Raises CalledProcessError on error.
+    """
+    with NamedNamespace(namespace) as ns:
+        routes = ns.check_output(["ip", "route"]).split("\n")
+        for route in routes:
+            if route.startswith("default") and "metric" not in route:
+                ns.check_output(["ip", "route", "del"] + route.split())
```
spikecurtis: Use `ip route change` here.

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +51,20 @@ def setup_logging(logfile, level=logging.INFO):
     _log.addHandler(handler)
 
 
+def increment_metrics(namespace):
+    """
+    Increment the metric of any default route that has a metric of 0.
+    :param namespace: The Networking namespace of the container.
+    :return: None. Raises CalledProcessError on error.
+    """
+    with NamedNamespace(namespace) as ns:
```
robbrockbank, line 66: Whilst we aren't trying to address all situations here, I think we could do better about there already being multiple default routes with different metrics.  I think the original design said something like:
-  Check if there is a default with metric 0, and if so increment all default metrics in reverse order.

For example: This gets round there being a default route with metric 0 and metric 1.

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +51,20 @@ def setup_logging(logfile, level=logging.INFO):
     _log.addHandler(handler)
 
 
+def increment_metrics(namespace):
+    """
+    Increment the metric of any default route that has a metric of 0.
```
spikecurtis: I thought @robbrockbank's design was to increment the metric of every default route if any have a metric of 0.  E.g. 
default via A metric 0
default via B metric 1

becomes

default via A metric 1
default via B metric 2

No?

calico_containers/pycalico/netns.py
spikecurtis, line 66: Hah!  Beat me to it.

calico_containers/pycalico/netns.py
robbrockbank: i think i just left a similar comment :-)

calico_containers/pycalico/netns.py
ozdanborne, line 66: Makes sense. but what's the importance of reverse order? Is it to prevent the algorithm from breaking when, during execution, a route is bumped to the same metric of anotther route? Since the routing table is snapshotted (`ns.check_output`), it should be fine, unless the routing table doesn't allow two routes with the same metric at any time?

robbrockbank, line 66: I believe it will not allow multiple default routes of the same metric.

calico_containers/pycalico/netns.py
ozdanborne: I toyed a lot with using `ip route change` to modify the metric, but couldn't get it to accept it:

```
/ # ip route
default via 172.17.42.2 dev eth0 
default via 172.17.42.1 dev eth0  metric 2
172.17.0.0/16 dev eth0  src 172.17.0.48 

/ # ip route change default via 172.17.42.2 dev eth0 metric 1
ip: RTNETLINK answers: No such file or directory

# But this goes through (without changing anything):
ip route change default via 172.17.42.2 dev eth0 metric 0
```

Note, this was on busybox, which almost seemed like it did not support `ip route change` as a function.

ozdanborne:
I ran this on a container that had 3 default routes, with metrics 0, 1, and 3. Seems to be working swell -

Before Calico:

```
/ # ip route
default via 172.17.42.2 dev eth0 
default via 172.17.42.1 dev eth0  metric 1 
default via 172.17.42.3 dev eth0  metric 3 
172.17.0.0/16 dev eth0  src 172.17.0.61 
```

after container-add:

```
/ # ip route
default via 172.24.114.18 dev eth1 
default via 172.17.42.2 dev eth0  metric 1 
default via 172.17.42.1 dev eth0  metric 2 
default via 172.17.42.3 dev eth0  metric 4 
172.17.0.0/16 dev eth0  src 172.17.0.61 
172.24.114.18 dev eth1 
```

After container-remove:

```
/ # ip route
default via 172.17.42.2 dev eth0  metric 1 
default via 172.17.42.1 dev eth0  metric 2 
default via 172.17.42.3 dev eth0  metric 4 
172.17.0.0/16 dev eth0  src 172.17.0.61 
```

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +52,55 @@ def setup_logging(logfile, level=logging.INFO):
...
+    :return: None. Raises CalledProcessError on error.
+    """
+    def get_metric(route):
+        """
+        Gets the metric of a route.
+        :param route: String representation of a route (full line returned by
+        `ip route`.
+        :return: Integer value representing the metric of the route.
+        """
+        match = re.search('metric\s+(\d)', route)
```
spikecurtis: doesn't `\d` match only a single digit?  Metrics are unsigned 32 bit numbers, so you'll handle multi-digit numbers.  Here and on line 96.

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +52,55 @@ def setup_logging(logfile, level=logging.INFO):
...
+            metric = 0
+        return metric
+
+    with NamedNamespace(namespace) as ns:
+        routes = ns.check_output(["ip", "route"]).split("\n")
+        default_routes = []
+        route_with_metric_0 = False
+        for route in routes:
+            if route.startswith("default"):
+                if "metric" not in route:
```
robbrockbank: I would rather you use the get_metric for this check.  More consistent.

Might have been easier to store in a dict keyed off metric and then you can just sort the keys to get an ordered list.

robbrockbank: better still, there is an OrderedDict class that might do some of that legwork...

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +52,55 @@ def setup_logging(logfile, level=logging.INFO):
...
+
+        if route_with_metric_0:
+            # Sort default routes by descending metric
+            descending_routes = sorted(default_routes,
+                                       lambda route_a, route_b:
+                                       cmp(get_metric(route_b),
+                                           get_metric(route_a)))
+            # Increment metric of each route by adding a new route,
+            # then removing the old route.
+            for route in descending_routes:
```
robbrockbank: Just a thought and possibly a complete overkill...  You could have had a "Route" class that is constructed from the route string.  That  way you could store the metric and the "rest" of the route incantation in one go, rather than having to do multiple string queries and then search/replaces.

spikecurtis: That's a good point.  The string you get back is structured data; we could just parse it.  Would also combine with other places in this module where we manipulate the routing table.

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +52,49 @@ def setup_logging(logfile, level=logging.INFO):
...
 
 
+def increment_metrics(namespace):
+    """
+    If any default route has a metric of 0, increase the metric of 
+    all default routes by 1.
+    :param namespace: The Networking namespace of the container.
+    :return: None. Raises CalledProcessError on error.
+    """
+    class Route(object):
```
robbrockbank: Will need UTs.

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +52,49 @@ def setup_logging(logfile, level=logging.INFO):
...
+
+    with NamedNamespace(namespace) as ns:
+        routes = ns.check_output(["ip", "route"]).split("\n")
+        default_routes = {}
+        for route in routes:
+            route = Route(route)
+            if route.is_default:
+                default_routes[route.metric] = route
+
+        if default_routes:
```
robbrockbank: i think this should be, `if 0 in default_routes`

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +52,49 @@ def setup_logging(logfile, level=logging.INFO):
...
+        def __init__(self, route_output):
+            self.route_output = route_output
+            self.default = route_output.startswith("default")
+            match = re.search('metric\s+(\d+)', route_output)
+            try:
+                self.metric = int(match.group(1))
+            except AttributeError:
+                self.metric = 0
+
+        def with_incremented_metric(self):
```
robbrockbank: don't forget the doc string :-)

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +52,49 @@ def setup_logging(logfile, level=logging.INFO):
...
+    :param namespace: The Networking namespace of the container.
+    :return: None. Raises CalledProcessError on error.
+    """
+    class Route(object):
+        def __init__(self, route_output):
+            self.route_output = route_output
+            self.default = route_output.startswith("default")
+            match = re.search('metric\s+(\d+)', route_output)
+            try:
+                self.metric = int(match.group(1))
```
robbrockbank: It's more traditional to check if match is None rather than trying it and catching an AttributeError.

Python has a ternary conditional operator too if you'd prefer to get this on one line:
`self.metric = int(match.group(1)) if match else 0`

calico_containers/pycalico/netns.py
```diff
@@ -209,6 +238,22 @@ def remove_ip_from_ns_veth(namespace, ip, veth_name_ns):
...
+        match = re.search('metric\s+(\d+)', route_output)
+        self.metric = int(match.group(1)) if match else 0
+
+    def with_incremented_metric(self):
+        """
+        Returns a string representing the route with its metric incremented
+        by 1. 
+        """
+        route_without_metric = re.sub('metric\s+\d+', '', self.route_output)
+        return "{} metric {}".format(route_without_metric, self.metric + 1)
```
spikecurtis: Think it's worth handling the maximum metric case (4294967295)?  E.g. `min(self.metric+1, MAX_METRIC)`?

robbrockbank: I was wondering the same.  We could just increment anything that can be incremented (i.e. if it is less than MAX and if the next highest metric won't clash).

This would then only be a problem if you had MAX default routes.  In this case all bets are off anyway.

ozdanborne: I see a couple of edge cases with that, and I'm not sure what lengths we want to go to to resolve them:
- if they have one route with `metric=max(metric)` and `metric=max(metric)-1`, we'd fail since the max would not be incremented, but max-1 would, and would fail since 2 routes can't have the same metric.

So it wouldn't be a problem if they had MAX default routes, rather it'd be a problem if they had a default route with metric 0, and a default route with metric=max

I think allowing the calls to fail and bubble up as exceptions is much simpler than trying to mitigate edge cases

spikecurtis: These are pretty crazy edge cases, so I'm happy with whatevs.

calico_containers/pycalico/netns.py
```diff
@@ -209,6 +251,18 @@ def remove_ip_from_ns_veth(namespace, ip, veth_name_ns):
...
 
+class Route(object):
+    def __init__(self, route_output):
+        self.route_output = route_output
+        self.default = route_output.startswith("default")
+        match = re.search('metric\s+(\d+)', route_output)
+        self.metric = int(match.group(1)) if match else 0
+
+    def __str__(self):
+        route_without_metric = re.sub('metric\s+\d+', '', self.route_output)
```
ozdanborne, line 263: I opted for simplicity by always specifiying the metric even if its 0, but I could do this:

```
route_without_metric = re.sub('metric\s+\d+', '', self.route_output)
if self.metric:
    return "{} metric {}".format(route_without_metric, self.metric)
else:
    return route_without_metric
```

I figure being explicit takes up less lines, so I just always print the metric. Up for suggestions though.

calico_containers/tests/unit/netns_test.py
```diff
@@ -85,3 +85,35 @@ def test_veth_exists_false(self, m_check_call, m_open):
...
+        route = Route("default via 172.24.114.1 dev eth0")
+        self.assertEqual(str(route), "default via 172.24.114.1 dev eth0 metric 0")
+
+        route.metric += 1
+        self.assertEqual(str(route),
+                         "default via 172.24.114.1 dev eth0 metric 1")
+
+        route.metric += 1
+        self.assertEqual(str(route),
+                        "default via 172.24.114.1 dev eth0 metric 2")
```
robbrockbank, line 120: I think you need to add one more test to test your MAX metric code.  Otherwise LGTM

calico_containers/pycalico/netns.py
```diff
@@ -51,6 +55,44 @@ def setup_logging(logfile, level=logging.INFO):
     _log.addHandler(handler)
 
 
+def increment_metrics(namespace):
+    """
+    If any default route has a metric of 0, increase the metric of
```
robbrockbank, line 60: The doc string is a little over simplified now :-)

robbrockbank:
Couple of minor comments that should be fixed, but I don't need to see it again.

LGTM once those are addressed.

robbrockbank:
LGTM

PR #44, opened 2015-10-24
Add a NullHandler to all the loggers

### LINKED ISSUE #42: libcalico should add a NullHandler as the default logger
Some code emits errors about not having a logger. Clients of this library can configure logging if they want, but they shouldn't get errors if they don't.

See https://github.com/projectcalico/calico-docker/issues/538

tomdee:
Fixes #42


robbrockbank:
LGTM

PR #45, opened 2015-10-29
Add function to check if interface exists

TrimBiggs:


calico_containers/pycalico/netns.py
```diff
@@ -132,6 +132,23 @@ def veth_exists(veth_name_host):
...
 
+def ns_veth_exists(namespace, veth_name_ns):
+    """
+    Check if veth exists in the container Namespace.
+    :param veth_name_ns: The name of the veth interface.
+    :return: True if veth exists, False if veth does not exist
+    """
+    # Suppress output
+    with NamedNamespace(namespace) as ns:
+        with open(os.devnull, 'w') as fnull:
```
robbrockbank: what's this for then?  you're not actually piping stdout or stdin into fnull so this isn't much use is it?

robbrockbank:
LGTM

PR #46, opened 2015-11-02
(DO NOT MERGE) Testing bulk information pulls for IP Addresses

insequent:
Hey All,

_TL;DR_:
My goal here is to make an efficient as possible pull for bulk data about IP Assignments.

_Reasoning_:
Calico's back-end is quite flexible, considering the number of front-ends (clients?) we have such as calico-docker, calico-mesos, etc. So far, I haven't seen any reason not to use multiple calico front-ends and/or other IPAM consumers against the ETCD datastore, barring the consistent use of the database driver, libcalico. However, with increased consumers, it would be nice to tell at a glance what is consuming what IPs. The most logically solution for this would be to include this data within the IP attributes.

Later on, Users could even include things like services, DC locations, etc, and the front-ends would be able to filter on those attributes. 

Currently however, I've only found the means to pull IP attributes is singular (`pycalico.ipam.IPAMClient.get_assignment_attributes`). This would result the number of requests against ETCD to equal the number of IPs in question, which could get quite large. Thus, the trade-off here being the number of request versus the size of the request. My sample code code included is designed to limit the number of requests to 2 + # of blocks that are inside the provided CIDR.

So please, take a look and comment as to whether this idea is totally insane and without merit, or if I should continue working on it and adding tests/validation.

Thanks!
Josh

### DIFF
--- a/calico_containers/pycalico/ipam.py
+++ b/calico_containers/pycalico/ipam.py
@@ -15,6 +15,7 @@
 from etcd import EtcdKeyNotFound, EtcdAlreadyExist, EtcdCompareFailed
 
 from netaddr import IPAddress, IPNetwork
+from netaddr.core import AddrFormatError
 import socket
 import logging
 import random
@@ -89,10 +90,18 @@ class BlockHandleReaderWriter(DatastoreClient):
         Get the blocks for which this host has affinity.
 
         :param host: The host name to get affinity for.
+        :type host: string
         :param version: 4 for IPv4, 6 for IPv6.
+        :type version: int
         :param pool: Limit blocks to a specific pool, or pass None to find all
         blocks for the specified version.
+        :type pool: IPPool
+        :return: A list of blocks with affinity to host
+        :rtype: list of IPNetworks
         """
+        # Validate before datastore request
+        assert isinstance(pool, IPPool) or None
+
         # Construct the path
         path = IPAM_HOST_AFFINITY_PATH % {"hostname": host,
                                           "version": version}
@@ -111,8 +120,35 @@ class BlockHandleReaderWriter(DatastoreClient):
 
         # If pool specified, filter to only include ones in the pool.
         if pool is not None:
-            assert isinstance(pool, IPPool)
             block_ids = [cidr for cidr in block_ids if cidr in pool]
+
+        return block_ids
+
+    def _get_all_blocks(self, cidr):
+        """
+        Get all blocks that are contained with a particular CIDR.
+
+        :param cidr: The IP Network whose blocks you need.
+        :type cidr: IPNetwork
+        :return: All blocks that lie within the given CIDR.
+        :rtype: list of IPNetworks
+        """
+        path = IPAM_BLOCK_PATH.format(version=cidr.version)
+
+        block_ids = []
+        try:
+            result = self.etcd_client.read(path).children or []
+        except EtcdKeyNotFound:
+            pass
+
+        for child in result:
+            block_id = child.key.split("/")[-1]
+            try:
+                block_ids.append(IPNetwork(block_id.replace("-", "/")))
+            except AddrFormatError:
+                pass
+
+        block_ids = [block_id for block_id in block_ids if block_id in cidr]
 
         return block_ids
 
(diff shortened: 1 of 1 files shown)


spikecurtis:
I can definitely see the value in being able to pull this data in bulk to use for diagnostic and operational reports.  I'll add a couple inline comments that will hopefully help steer your implementation.

calico_containers/pycalico/ipam.py
```diff
@@ -782,6 +820,56 @@ def get_assignment_attributes(self, address):
...
+        pools = self.get_ip_pools(cidr.version)
+
+        result = []
+        for block in blocks:
+            for pool_candidate in pools:
+                if block in pool_candidate:
+                    pool = str(pool_candidate)
+                    break
+
+            for num, index in enumerate(block.allocations):
```
spikecurtis, line 854: I suggest you move this code into block.py.  It works with internals of the `AllocationBlock` class and is nice and self contained to a single block.  That way we keep the interface to 'AllocationBlock` abstract from its implementation.

calico_containers/pycalico/ipam.py
```diff
@@ -782,6 +820,56 @@ def get_assignment_attributes(self, address):
...
+                ip_int = block.first + num
```
spikecurtis, line 855: You probably want `block.cidr.first` here.

calico_containers/pycalico/ipam.py
```diff
@@ -782,6 +820,56 @@ def get_assignment_attributes(self, address):
...
+                    if index is None:
+                        ip_attrs["ALLOCATED"] = False
+                        ip_attrs["ATTRIBUTES"] = {}
+                        ip_attrs["HANDLE"] = None
+                    else:
+                        attributes = block.attributes[index]
+                        ip_attrs["ATTRIBUTES"] = attributes["secondary"]
+                        ip_attrs["HANDLE"] = attributes["handle_id"]
+                        ip_attrs["IP"] = str(IPAddress(ip_int))
+                    result.append(ip_attrs)
```
spikecurtis, line 868: The encoding on the wire is designed for the case where many IP addresses share the same set of attributes & handle.  Joining the data to have each unique IP in a list might make the result structure very large.  We should think carefully about downstream applications consuming this data---what is a useful format for them?

Use case you've mentioned: report number of IPs with particular attributes (e.g. "orchestrator": "mesos").  

Any others?

calico_containers/pycalico/ipam.py
```diff
@@ -89,10 +90,18 @@ def _get_affine_blocks(self, host, version, pool):
...
         :param version: 4 for IPv4, 6 for IPv6.
+        :type version: int
         :param pool: Limit blocks to a specific pool, or pass None to find all
         blocks for the specified version.
+        :type pool: IPPool
+        :return: A list of blocks with affinity to host
+        :rtype: list of IPNetworks
         """
+        # Validate before datastore request
+        assert isinstance(pool, IPPool) or None
```
insequent, line 103: this should be `or pool is None`

spikecurtis, line 103: agree

calico_containers/pycalico/ipam.py
insequent, line 868: The size of this data could certainly get quite large. I was debating whether it would be better to leave the objects as is (i.e. pool = IPPool,  ip = IPAddress), or converting them to strings was the better option. Converting them back for the consumer could get expensive, but if we leave as is, the result is large.

Last night I was thinking strings would be cleaner, because my original intent was User read reports, however this morning I was thinking that the data transfer of these large results is already done. I.e. the damage is already done, memory is allocated and the datastore has already been hit, so why not return the whole objects and let the consumer decide what to do with it.

The only other use case that comes to mind, is a return bulk update operation. The consumer could use this get the objects they need, update them, then send them back.

insequent:
Thanks for sanity check, @spikecurtis! I'll make the suggested changes and start in on tests, hopefully later tonight.

insequent:
And by tonight, I meant 3 days later :/. Work keeping me busy.

EDIT: Still busy. I plan to attack this ASAP, but keep getting bogged down with other things. Got 2 high priorities atm, but should be able to get back to this soon (hopefully Tuesday night).

EDIT: (11/16/15) My apologies, but time really slipped away from me for a while there. I'll be working on this over the next week.

insequent:
Sorry for the delay, but work has me literally working until midnight every night. I likely won't be able to wrap this up until Sunday.

insequent:
Again, sorry for my fantastic amount of slacking. I'll for sure be working on this over the next week.

insequent:
Going to close this for now until I find the time for this. I've been piecemealing it, but not making real headway. Need a quiet day to just work through it. Hopefully I'll be back soon.

Issue #47, opened 2015-11-02
IPAM raising KeyError


plwhite:
It's not impossible that this is a problem with calling code, but here is a stack trace that to me implies a possible bug in the libcalico IPAM.

My suspicion is that the issue was caused by multiple hosts attempting to grab an IP at the same time, and that one of them found that the block it was about to grab went away under its feet (this is a moderately sized scale test with 200 hosts, one of which failed). Is that plausible? If so, I'd like to aim for a workaround (assuming the fix isn't too hard). Would catching the KeyError and trying again be sensible?

```
2015-11-02 15:07:47,642:INFO:auto_assign_ips(407): Auto-assign 1 IPv4, 0 IPv6 addrs
2015-11-02 15:07:47,673:INFO:_auto_assign(457): Ran out of affine blocks for plw-host-0138 in pool None
2015-11-02 15:07:48,427:ERROR:<module>(405): Main exiting
Traceback (most recent call last):
  File "./host-agent.py", line 403, in <module>
    main()
  File "./host-agent.py", line 386, in main
    start_job(job_id, sub_job, etcd_client)
  File "./host-agent.py", line 130, in start_job
    attributes={})
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 409, in auto_assign_ips
    attributes, pool[0], hostname)
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 475, in _auto_assign
    pool)
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 151, in _new_affine_block
    self._claim_block_affinity(host, block_cidr)
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 177, in _claim_block_affinity
    block = self._read_block(block_cidr)
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 65, in _read_block
    raise KeyError(str(block_cidr))
KeyError: '192.168.56.0/26'
```

The calling code in `host-agent.py` looks like this.

```
            ip_list, _ = ipam_client.auto_assign_ips(num_v4=1,
                                                     num_v6=0,
                                                     handle_id=None,
                                                     attributes={})
```

spikecurtis:
Ung.  There shouldn't be any issue with blocks disappearing under anyone's feet for the simple reason that _we never delete blocks_---only create and update.

[...]

PR #49, opened 2015-11-04
Use quorum=True for etcd reads in IPAM.

spikecurtis:
Fixes https://github.com/projectcalico/libcalico/issues/47

### DIFF
--- a/calico_containers/pycalico/ipam.py
+++ b/calico_containers/pycalico/ipam.py
@@ -29,7 +29,8 @@ from pycalico.block import (AllocationBlock,
                             get_block_cidr_for_address,
                             BLOCK_PREFIXLEN,
                             AlreadyAssignedError,
-                            AddressNotAssignedError)
+                            AddressNotAssignedError,
+                            NoHostAffinityWarning)
 from pycalico.handle import (AllocationHandle,
                              AddressCountTooLow)
 from pycalico.util import get_hostname
@@ -38,6 +39,8 @@ _log = logging.getLogger(__name__)
 _log.addHandler(logging.NullHandler())
 
 RETRIES = 100
+
+KEY_ERROR_RETRIES = 3
 
 
 class BlockHandleReaderWriter(DatastoreClient):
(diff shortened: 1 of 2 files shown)


calico_containers/pycalico/ipam.py
```diff
@@ -455,10 +459,18 @@ def _auto_assign(self, ip_version, num, handle_id,
...
+                ips = self._auto_assign_block(block_id,
+                                              num_remaining,
+                                              handle_id,
+                                              attributes)
+            except (KeyError, NoHostAffinityWarning):
+                # In certain rare race conditions, _get_affine_blocks above
+                # can return block_ids that don't exist or don't actually have
+                # affinity to this host (due to multiple IPAM clients on this
+                # host running simultaneously).  If that happens, just move to
+                # the next one.
```
robbrockbank: In the event of a KeyError don't we want to retry from the beginning again?  i.e. get the affine blocks again and then retry (but perhaps skipping the blocks we already tried)?  Otherwise we potentially assign another affine block when it isn't necessary.

calico_containers/pycalico/ipam.py
```diff
@@ -455,10 +459,18 @@ def _auto_assign(self, ip_version, num, handle_id,
                 _log.info("Ran out of affine blocks for %s in pool %s",
                           hostname, pool)
                 break
-            ips = self._auto_assign_block(block_id,
-                                          num_remaining,
-                                          handle_id,
-                                          attributes)
+            try:
+                ips = self._auto_assign_block(block_id,
```
robbrockbank, line 466: If this is the last block and you get a KeyError (because the client creating this block hasn't written it yet) - shouldn't we be retrying rather than potentially creating a new affine block below.

spikecurtis, line 466: Done!

calico_containers/pycalico/ipam.py
```diff
@@ -444,21 +450,49 @@ def _auto_assign(self, ip_version, num, handle_id,
...
+        block_ids.reverse()
+        key_errors = 0
         allocated_ips = []
 
         num_remaining = num
         while num_remaining > 0:
             try:
-                block_id = block_ids.next()
-            except StopIteration:
+                block_id = block_ids.pop()
```
robbrockbank: pop(0) could have been used instead of needing to reverse the list - but not fussed either way

spikecurtis: Today I Learned...

robbrockbank:
LGTM

PR #50, opened 2015-11-06
Add support for secure etcd

TrimBiggs:
This change requires updates to the calico-docker and calico repos.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -147,25 +147,24 @@ class DatastoreClient(object):
                                                 ETCD_KEY_FILE_ENV, etcd_key,
                                                 ETCD_CERT_FILE_ENV, etcd_cert))
             # Make sure etcd key and certificate are readable
-            if etcd_key and etcd_cert:
-                if not (os.path.isfile(etcd_key) and
-                        os.access(etcd_key, os.R_OK) and
-                        os.path.isfile(etcd_cert) and
-                        os.access(etcd_cert, os.R_OK)):
-                    raise DataStoreError("Cannot read %s and/or %s. Both must "
-                                         "be readable file paths. Values "
-                                         "provided: %s=%s, %s=%s" %
-                                         (ETCD_KEY_FILE_ENV,
-                                          ETCD_CERT_FILE_ENV,
-                                          ETCD_KEY_FILE_ENV, etcd_key,
-                                          ETCD_CERT_FILE_ENV, etcd_cert))
-                # If Certificate Authority cert provided, check it's readable
-                if etcd_ca and not (os.path.isfile(etcd_ca) and
-                                    os.access(etcd_ca, os.R_OK)):
-                    raise DataStoreError("Cannot read %s. Value must be "
-                                         "readable file path. Value provided: "
-                                         "%s" % (ETCD_CA_CERT_FILE_ENV,
-                                                 etcd_ca))
+            if etcd_key and etcd_cert and not (os.path.isfile(etcd_key) and
+                                               os.access(etcd_key, os.R_OK) and
+                                               os.path.isfile(etcd_cert) and
+                                               os.access(etcd_cert, os.R_OK)):
+                raise DataStoreError("Cannot read %s and/or %s. Both must "
+                                     "be readable file paths. Values "
+                                     "provided: %s=%s, %s=%s" %
+                                     (ETCD_KEY_FILE_ENV,
+                                      ETCD_CERT_FILE_ENV,
+                                      ETCD_KEY_FILE_ENV, etcd_key,
+                                      ETCD_CERT_FILE_ENV, etcd_cert))
+            # Certificate Authority cert must be provided, check it's readable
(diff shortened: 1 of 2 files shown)


calico_containers/pycalico/datastore.py
```diff
@@ -29,6 +29,16 @@
 ETCD_AUTHORITY_DEFAULT = "127.0.0.1:4001"
 ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
 
+# Secure etcd with SSL environment variables and paths
+ETCD_SCHEME_DEFAULT = "http"
+ETCD_SCHEME_ENV = "ETCD_SCHEME_ENV"
+ETCD_KEY_FILE_ENV = "ETCD_KEY_FILE"
+ETCD_KEY_NODE_FILE = "/etc/calico/ssl/key.pem"
```
robbrockbank: These node variables do not belong in libcalico - they are calico-docker specific.

calico_containers/pycalico/datastore.py
```diff
@@ -29,6 +29,16 @@
 ETCD_AUTHORITY_DEFAULT = "127.0.0.1:4001"
 ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
 
+# Secure etcd with SSL environment variables and paths
+ETCD_SCHEME_DEFAULT = "http"
+ETCD_SCHEME_ENV = "ETCD_SCHEME_ENV"
```
robbrockbank: No _ENV on the actual environment variable name

calico_containers/pycalico/datastore.py
```diff
@@ -113,6 +123,51 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
+def assert_valid_etcd_keys(scheme, key, cert, ca_cert):
```
robbrockbank: Strictly speaking this is not asserting (which would be if it raised an AssertionError).  Probably should be called validate_etcd_keys or something similar.

Also, the interface and messages don't really marry correctly.  The interface has you pass in the keys and certificates, but the error messages describe the environment variable names.

I wonder - should this simply be part of the Datastore connection processing rather than an external helper method?

calico_containers/pycalico/datastore.py
```diff
@@ -113,6 +123,51 @@ def wrapped(*args, **kwargs):
...
+                                                       ETCD_KEY_FILE_ENV,
+                                                       key,
+                                                       ETCD_CERT_FILE_ENV,
+                                                       cert))
+            # If Certificate Authority cert file provided, check it's readable
+            if ca_cert and not os.access(ca_cert, os.R_OK):
+                raise DataStoreError("Cannot read %s. Value must be readable "
+                                     "file path. Value provided: %s" %
+                                     (ETCD_CA_CERT_FILE_ENV, ca_cert))
+    elif not (scheme == "http" or scheme == ""):
```
robbrockbank: Depends on what this is actually checking, but at this point I'd expect scheme to be https or http and not blank (i.e. we would have applied our default at this point).  I guess I'm wondering if the API for this method should allow a blank value of this.

TrimBiggs: I was wondering this as well.  I went back and forth with it a few times.  I think this is specific enough that it should only be used in this one case though, so checking this against "http" alone should be enough.

calico_containers/pycalico/datastore.py
TrimBiggs: Woops! :stuck_out_tongue_closed_eyes:

calico_containers/tests/unit/datastore_test.py
```diff
@@ -81,6 +84,14 @@
                  "active", "11-22-33-44-55-66")
 EP_12.profile_ids = ["UNIT"]
 
+ETCD_ENV_DICT = {
```
robbrockbank, line 87: Very minor point, I think you only reference this once, so perhaps just as easy to embed in the one place that uses it?

robbrockbank:
LGTM

This looks good.  And just to check, we have ST'd the different permutations of cert, key, ca and scheme to make sure it does actually work?

TrimBiggs:
Yes, I tested both security methods:
1) etcd Server has (key, cert) pair, Client accesses with CA cert
2) etcd Server has (key1, cert1) pair and CA2 cert, Client accesses with (key2, cert2) and CA1 cert

calico_containers/pycalico/datastore.py
```diff
@@ -25,10 +25,22 @@
 from pycalico.datastore_errors import DataStoreError, \
     ProfileNotInEndpoint, ProfileAlreadyInEndpoint, MultipleEndpointsMatch
 from pycalico.util import get_hostname
+import logging
+
+# Suppress warnings from urllib3 when etcd is used without CA cert.
+logging.captureWarnings(True)
```
robbrockbank: shouldn't you restrict this to urllib3?

robbrockbank: Actually i think this should be in calicoctl rather than here.  Also feel free to add logs to your branches, but you'll need to add a null handler for this logger.

PR #51, opened 2015-11-13
Added mtu keyword arg for optional MTU when adding a veth

TrimBiggs:


TrimBiggs:
This PR coincides with https://github.com/projectcalico/calico-docker/pull/624 to fix https://github.com/projectcalico/calico-docker/issues/610

TrimBiggs:
This isn't how we want to fix the issue.

PR #52, opened 2015-11-17
Add flag to IP Pools to allow pool to be excluded from Calico IPAM assignment

robbrockbank:


calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -156,14 +161,16 @@ def from_json(cls, json_str):
         json_dict = json.loads(json_str)
         return cls(json_dict["cidr"],
                    ipip=json_dict.get("ipip"),
-                   masquerade=json_dict.get("masquerade"))
+                   masquerade=json_dict.get("masquerade", False),
```
tomdee: Why this change to add the default?

calico_containers/pycalico/datastore.py
```diff
@@ -326,11 +326,13 @@ def get_host_bgp_ips(self, hostname):
             return (ipv4, ipv6)
 
     @handle_errors
-    def get_ip_pools(self, version):
+    def get_ip_pools(self, version, ipam_only=False):
```
tomdee: Why have ipam only in one place and skip ipam in another? Would it be better to be consistent?

tomdee:
Couple of minor comments but otherwise LGTM

calico_containers/pycalico/datastore_datatypes.py
robbrockbank: No reason whatsoever!  I'll remove.

calico_containers/pycalico/datastore.py
robbrockbank: Yeah I will change that!  It was originally because I wanted a lack of value in the JSON to be "False", but that really isn't necessary.

PR #53, opened 2015-11-17
Validate args for creating a Rule

luke-mino-altherr:


calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -396,6 +396,14 @@ def __setitem__(self, key, value):
             value = IPNetwork(value)
         if key == "action" and value not in ("allow", "deny"):
             raise ValueError("'%s' is not allowed for key 'action'" % value)
+        if key == "protocol" and value not in ("tcp", "udp", "icmp", None):
+            raise ValueError("'%s' is not allowed for key 'protocol'" % value)
+        if key in ("src_tag", "dst_tag") and re.search('[^a-zA-Z0-9\.\_\-]', value):
+            raise ValueError("'%s' is not allowed for key '%s'" % (value, key))
+        if key in ("src_ports", "dst_ports") and not check_ports(value):
+            raise ValueError("'%s' is not allowed for key '%s'" % (value, key))
+        if key == "icmp_type" and value not in range(256):
```
robbrockbank: In Python2 I think range() creates a list, so this isn't very efficient.  Correct me if I'm wrong tho.

calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -396,6 +396,14 @@ def __setitem__(self, key, value):
             value = IPNetwork(value)
         if key == "action" and value not in ("allow", "deny"):
             raise ValueError("'%s' is not allowed for key 'action'" % value)
+        if key == "protocol" and value not in ("tcp", "udp", "icmp", None):
+            raise ValueError("'%s' is not allowed for key 'protocol'" % value)
+        if key in ("src_tag", "dst_tag") and re.search('[^a-zA-Z0-9\.\_\-]', value):
```
robbrockbank: You don't need to escape any of the last characters do you?
. should mean "." when in [ ]
I don't think _ means anything
and the - should be fine when at the beginning or the end of the list of chars

might be wrong tho.

calico_containers/pycalico/util.py
```diff
@@ -90,4 +90,24 @@ def get_hostname():
...
+    Checks whether a list of ports are within range of 0 and 65535.
+    The port list may contains ints or strings including a a range of ports,
+    delimited by a colon.
+
+    :param port_list:
+    :return: a Boolean: True if in range, False if not in range
+    """
+    in_range = True
+    for port in port_list:
+        if ":" in str(port):
```
robbrockbank, line 109: Shouldn't 1:2:3 be invalid?  I think this will mark is as valid though.

Also shouldn;'t the second be higher than the first?

calico_containers/pycalico/util.py
```diff
@@ -90,4 +90,24 @@ def get_hostname():
...
+            in_range = check_ports(port.split(":"))
+        else:
+            in_range = 0 <= int(port) < 65536
```
robbrockbank: this will raise an exception if the port value is not a number.  That may be correct, but should make that clear in the doc string.

calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -396,6 +396,14 @@ def __setitem__(self, key, value):
...
+            raise ValueError("'%s' is not allowed for key '%s'" % (value, key))
+        if key in ("src_ports", "dst_ports") and not check_ports(value):
+            raise ValueError("'%s' is not allowed for key '%s'" % (value, key))
+        if key == "icmp_type" and value not in range(256):
+            raise ValueError("'%s' is not allowed for key '%s'" % (value, key))
```
robbrockbank, line 407: How do these errors manifest themselves when using calicoctl?  Do they make sense, or do we need to do some additional checking in calicoctl?

robbrockbank:
We do a lot of this checking in calicoctl (see profile.py and utils.py).  It strikes me that we should probably remove some of the verification methods from calicoctl into libcalico and then update calicoctl to make use of them.

robbrockbank:
LGTM - can you squash before merging?

Issue #57, opened 2015-11-23
The hostname argument for auto_assign_ips is not respected


plwhite:
When you call `auto_assign_ips`,  there is an optional `hostname` argument (defaulting to the hostname of this server). In principle, that means that you can have a name for a server used in IPAM that is not the same as the server hostname. Unfortunately, if you don't set it to the hostname of this server, then it doesn't work, which makes it a bit pointless.

The underlying issue is that `_auto_assign_block` is called, and that raises an exception if the block in question does not match the server hostname (not the `hostname` parameter). One fix would be to remove that check, but that seems a bit crazy (it's a valid check). Another fix would be to pass the supplied hostname down to `_auto_assign_block`, which I think is probably the correct fix.

I can probably put together a PR for this, but it'd be good to get confirmation that you agree with my analysis in case I've misunderstood how to use the interface.

Here's a stack trace of what happened when I tried this (`dummy-0000` is the hostname of the box, not what I passed, and this is build 0.3.0) : 

```
Traceback (most recent call last):
  File "./host-agent.py", line 532, in <module>
    main()
  File "./host-agent.py", line 374, in main
    do_ipam(name)
  File "./host-agent.py", line 342, in do_ipam
    hostname=name)
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 409, in auto_assign_ips   
    attributes, pool[0], hostname)
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 484, in _auto_assign
    attributes)
  File "/usr/local/lib/python2.7/site-packages/pycalico/ipam.py", line 538, in _auto_assign_block
    affinity_check=affinity_check)
  File "/usr/local/lib/python2.7/site-packages/pycalico/block.py", line 169, in auto_assign
    self.host_affinity)
pycalico.block.NoHostAffinityWarning: Host affinity is dummy-0000
```

robbrockbank:
@plwhite 
Yup - looks like the correct fix is to plumb through the hostname.  Feel free to assign a PR to me if you get a fix out for this.

spikecurtis:
@robbrockbank, @tomdee and I discussed this at length and our conclusion was to use the environment variable exclusively, rather than as a parameter on the function call.  If we're changing this again, I'd like us to reach a consistent position across the entire `libcalico` API.

robbrockbank:
[...]

PR #61, opened 2015-12-10
Rename hostname to be host and fix use of hostname parm in API

robbrockbank:
Allow host to be passed in on assign ip methods.

### DIFF
--- a/calico_containers/pycalico/ipam.py
+++ b/calico_containers/pycalico/ipam.py
@@ -15,7 +15,6 @@
 from etcd import EtcdKeyNotFound, EtcdAlreadyExist, EtcdCompareFailed
 
 from netaddr import IPAddress, IPNetwork
-import socket
 import logging
 import random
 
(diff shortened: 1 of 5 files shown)


robbrockbank:
Fixes #57

calico_containers/pycalico/ipam.py
```diff
@@ -818,62 +819,3 @@ def get_assignment_attributes(self, address):
         else:
             _, attributes = block.get_attributes_for_ip(address)
             return attributes
-
```
tomdee, line 821: We;ll have to make sure this is mentioned in the release note.

tomdee:
Can you confirm that other than the variable renaming and test updates, the only changes are
- add host to auto_assign (+ update the affinity check)
- add host to _auto_assign_block (+ pass through)
- remove a couple of depracated methods?

robbrockbank:
Yes that's it + curresponding text reworking to talk about host ID rather than hostname

tomdee:
OK, so if a user consistently passes in their own host parameter, they could use whatever values they want and get_hostname() will never be called?

If so, LGTM.

robbrockbank:
@tomdee, yes that's correct.

Merging.

calico_containers/pycalico/ipam.py
robbrockbank, line 821: Yup - will do

PR #63, opened 2015-12-17
Add method to get hostname from BGP IP in etcd

TrimBiggs:


calico_containers/pycalico/datastore.py
```diff
@@ -306,6 +310,31 @@ def remove_host(self, hostname):
...
+
+        ip_host_dict = {}
+
+        # Loop through key-value pairs to find IP addresses
+        for host_ip in host_ips:
+            if host_ip.value in ip_list:
+                # Pull the hostname from the datastore key string
+                hostname = HOSTNAME_DATASTORE_RE.match(host_ip.key).group(1)
+                ip_host_dict[host_ip.value] = hostname
+        # No hosts are associated with the given IP
```
robbrockbank: Should this comment be here?

calico_containers/pycalico/datastore.py
```diff
@@ -306,6 +310,31 @@ def remove_host(self, hostname):
...
+            host_ips = hosts.children
+        except EtcdKeyNotFound:
+            # No BGP hosts currently configured in etcd, so no host owns the IP
+            raise KeyError("No BGP host configurations found.")
+
+        ip_host_dict = {}
+
+        # Loop through key-value pairs to find IP addresses
+        for host_ip in host_ips:
+            if host_ip.value in ip_list:
```
robbrockbank: Surely you need to match the key before matching the value?  A recursive read will return more than just the host BGP IP bindings.

calico_containers/pycalico/datastore.py
```diff
@@ -306,6 +310,31 @@ def remove_host(self, hostname):
...
     @handle_errors
+    def get_hostnames_from_ips(self, ip_list):
+        """
+        Get the hostnames that are using the given IPs as their calico node IPs.
+        :param ip_list: The list of IPs to get hostnames for.
+        :return: A dictionary of {IP:hostname} the hosts that own the given IPs.
+        """
+        try:
+            hosts = self.etcd_client.read(BGP_HOSTS_PATH, recursive=True)
+            host_ips = hosts.children
```
robbrockbank: Should this be leaves rather than children?

TrimBiggs: Yup, you're right. Children is deprecated and just returns leaves.

robbrockbank:
LGTM

PR #65, opened 2015-12-18
Raise PoolNotFound rather than ValueError

robbrockbank:


spikecurtis:
Code looks fine, but I'd like to understand the context a bit better.  @robbrockbank call me when you get in.

PR #69, opened 2016-01-04
Release by handle should update the unallocated array

robbrockbank:
Spike - spotted this while doing the IPAM stuff - I wasn't setting the unallocated entry in the release by handle method.


spikecurtis:
Can you add a test to `test_block.py` (or add some assertions to an existing test of `release_by_handle()`)?

spikecurtis:
LGTM!

PR #70, opened 2016-01-04
Allow icmpv6 profile rule option

TrimBiggs:


calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -391,6 +391,8 @@ class Rule(dict):
                     "dst_net",
                     "icmp_type",
                     "icmp_code",
+                    "icmpv6_type",
```
robbrockbank: I don't think this is correct.

The source for calico suggests that for ICMP and ICMPv6 the code and type field are just called icmp_code and icmp_type.

calico_containers/pycalico/util.py
```diff
@@ -142,7 +142,7 @@ def validate_icmp_type(icmp_type):
     Validate that icmp_type is an integer between 0 and 255.
     If not return False.
 
-    :param icmp_type:
+    :param icmp_type: string or number value representing an icmp/icmpv6 type
```
robbrockbank: can it really be a string?  if so is it just a str(<int>) or do we expect some named values.  Do we have any tests that it can be a string?

TrimBiggs: We have some tests in calicoctl where it is a stringified number like "100", but I guess that doesn't have to be explicit here.

robbrockbank:
LGTM - can you squash before merging.

PR #74, opened 2016-01-07
Pre assign blocks

robbrockbank:


calico_containers/pycalico/block.py
```diff
@@ -179,18 +200,20 @@ def auto_assign(self, num, handle_id, attributes, host,
         :param attributes: Contents of this dict will be stored with the
         assignment and can be queried using get_assignment_attributes().  Must
         be JSON serializable.
-        :param host: The host ID to use for affinity in when assigning IP
-        addresses.
+        :param host: The ID of the host requesting addresses.
         :param affinity_check: If true, verify that this block's affinity
-        matches the supplied host and throw a NoHostAffinityWarning if it
-        doesn't.  Set to false to disable this check.
+        matches the supplied host and throw a NoHostAffinityError if it
```
caseydavenport: Is ignoring it the correct behavior? Or should we be raising an error indicating that the calling code is doing something wrong?

caseydavenport: Actually, scratch that, this is the right thing to do I think.

calico_containers/pycalico/block.py
```diff
@@ -179,18 +200,20 @@ def auto_assign(self, num, handle_id, attributes, host,
...
+        strict_host_affinity set to True, this parameter is ignored, and an
+        affinity check is always performed.
         :return: List of assigned addresses.  When the block is at or near
         full, this method may return fewer than requested IPs.
         """
         assert num >= 0
 
-        if affinity_check and host != self.host_affinity:
-            raise NoHostAffinityWarning("Block host affinity is %s (not %s)" %
+        check_affinity = self.strict_host_affinity or affinity_check
```
caseydavenport: You don't need to declare a new variable here.

```
affinity_check = self.strict_host_affinity or affinity_check
```

calico_containers/pycalico/block.py
```diff
@@ -214,10 +237,12 @@ def auto_assign(self, num, handle_id, attributes, host,
                 ips.append(ip)
         return ips
 
-    def assign(self, address, handle_id, attributes):
+    def assign(self, address, handle_id, attributes, host):
```
caseydavenport: Shouldn't `host` be an optional parameter? It is only required when `self.strict_host_affinity == True`

calico_containers/pycalico/block.py
```diff
@@ -214,10 +237,12 @@ def auto_assign(self, num, handle_id, attributes, host,
...
         """
```
caseydavenport: Also this method has a redundant `return` at the end of it.

calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -492,3 +492,75 @@ def pprint(self):
...
+    """
+    IPAM configuration.
+    """
+    AUTO_ALLOCATE_BLOCKS = "auto_allocate_blocks"
+    STRICT_HOST_AFFINITY = "strict_host_affinity"
+
+    def __init__(self, auto_allocate_blocks=True, strict_host_affinity=False):
+        self.auto_allocate_blocks = auto_allocate_blocks
+        """
+        Whether Calico IPAM module is allowed to auto-allocated affine blocks
```
caseydavenport: Typo "auto-allocated"

calico_containers/pycalico/block.py
```diff
@@ -67,14 +68,24 @@ def __init__(self, cidr_prefix, host_affinity):
         self.cidr = cidr_prefix
         self.db_result = None
 
-        self.host_affinity = host_affinity
+        self.host_affinity = host_affinity or None
```
caseydavenport: `host_affinity` is a required argument, no?  If the user of this class wants to disable host affinity, they need to explicitly pass `None`, so I think we don't need this "or" clause here.

calico_containers/pycalico/block.py
```diff
@@ -112,10 +123,12 @@ def to_json(self):
         """
         Convert to a JSON representation for writing to etcd.
         """
+        # Convert a host value of None to an empty string.
+        host = self.host_affinity or ""
 
         json_dict = {AllocationBlock.CIDR: str(self.cidr),
                      AllocationBlock.AFFINITY:
-                         AllocationBlock.HOST_AFFINITY_T % self.host_affinity,
+                         AllocationBlock.HOST_AFFINITY_T % host,
```
caseydavenport: Is this the correct way to express no affinity?

Would

```
{
  ... 
  affinity: "",
  ...
}
```

make more sense?

I think it does - it doesn't have any other sort of affinity either.

calico_containers/pycalico/block.py
```diff
@@ -129,12 +142,20 @@ def from_etcd_result(cls, etcd_result):
...
+        # specific host affinity.
         affinity = json_dict[AllocationBlock.AFFINITY]
         assert affinity[:5] == "host:"
-        host_affinity = affinity[5:]
+        host_affinity = affinity[5:] or None
 
-        block = cls(cidr_prefix, host_affinity)
+        # Parse out the strict_host_affinity flag.  If this does not exist
+        # assume False.
+        strict_host_affinity = json_dict.get(
```
caseydavenport: Should this field actually be called `strict_host_affinity` or should it be called `strict_affinity`?

Even though our IPAM code only supports host affinity currently, I believe the data model has been written to support various other types (e.g rack) affinity as well.

calico_containers/pycalico/block.py
```diff
@@ -516,6 +556,16 @@ def get_block_cidr_for_address(address):
     return IPNetwork(block_id)
 
 
+def validate_block_size(cidr):
+    """
+    Check that the CIDR block size is valid.  This checks that it is at least
+    as large as the minimum block size.
+    """
+    assert isinstance(cidr, IPNetwork)
+    #TODO Should probably limit the maximum block size too.
```
caseydavenport: Are we planning to do this TODO in this PR?

If not we should probably remove the TODO and just raise an issue.

calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -492,3 +492,75 @@ def pprint(self):
...
+        """
+        return {
+            IPAMConfig.AUTO_ALLOCATE_BLOCKS: self.auto_allocate_blocks,
+            IPAMConfig.STRICT_HOST_AFFINITY: self.strict_host_affinity
+        }
+
+    @classmethod
+    def from_json(cls, json_str):
+        """
+        Create an IPAMConfiguration from the raw JSON.
```
caseydavenport: IPAMConfiguration -> IPAMConfig

calico_containers/pycalico/ipam.py
```diff
@@ -92,7 +90,17 @@ def _compare_and_swap_block(self, block):
...
-    @handle_errors
+    def _delete_block(self, block):
+        """
+        Delete a block using an atomic delete operation.
+        """
+        try:
+            self.etcd_client.delete(
+                block.db_result.key,
+                prevIndex=block.db_result.modifiedIndex)
+        except EtcdAlreadyExist:
```
caseydavenport: Is `EtcdAlreadyExist` actually the exception that will get thrown by `delete`?

calico_containers/pycalico/block.py
robbrockbank: yes i think that makes more sense

calico_containers/pycalico/block.py
robbrockbank: yeah i think that makes more sense given our current model.  i'll update the design doc and the code

calico_containers/pycalico/block.py
robbrockbank: yeah - not sure why i did.

calico_containers/pycalico/ipam.py
```diff
@@ -199,13 +208,61 @@ def _claim_block_affinity(self, host, block_cidr):
...
+                    # The block is empty, so just delete the block.
+                    _log.debug("Block is empty - delete block")
+                    self._delete_block(block)
+                elif block.strict_host_affinity:
+                    # The block has strict host affinity, therefore it can
+                    # only contain IPs belonging to the host, so it is safe to
+                    # remove the block.
+                    _log.debug("Block only contains hosts IPs - delete block")
+                    self._delete_block(block)
+                else:
```
caseydavenport, line 281: This branch could use a comment like the ones above.

calico_containers/pycalico/block.py
robbrockbank, line 244: i think that makes for a yuckier interface since it requires the user to check that parameter before deciding whether to pass in the host.  

As an aside, and further afield we might need to pass in a bunch of different affinity values (rack, host etc ... ), but since this interface is used internally by the IPAM client, I don't think it is worth designing that at this point.

caseydavenport, line 244: Yep, sounds fine to me.

I was just having similar thoughts myself and was about to delete the comment :)

calico_containers/pycalico/ipam.py
```diff
@@ -199,13 +208,61 @@ def _claim_block_affinity(self, host, block_cidr):
...
+                          "Block %s is claimed by %s",
+                          block_cidr, block.host_affinity)
+
+            try:
+                if block.is_empty():
+                    # The block is empty, so just delete the block.
+                    _log.debug("Block is empty - delete block")
+                    self._delete_block(block)
+                elif block.strict_host_affinity:
+                    # The block has strict host affinity, therefore it can
```
caseydavenport: Hm.

Is this right?  I'm not sure I fully understand the reasoning / use-case here (or I've just forgotten it).

If we've assigned a few IPs from this block (all on the same host) and then release affinity on this block, those endpoints may still have those IPs, and those IPs might be re-assigned to another host some time in the future.  Is this OK? We could end up with multiple endpoints with the same IP.

Feels like this branch should do the same as the one below, but perhaps I'm missing something.

calico_containers/pycalico/ipam.py
```diff
@@ -349,6 +404,69 @@ def _compare_and_swap_handle(self, handle):
...
+        :param config: An IPAMConfig object.
+        """
+        assert isinstance(config, IPAMConfig)
+        current = self.get_ipam_config()
+        if current == config:
+            _log.debug("Configuration has not changed")
+            return
+
+        blocksv4, blocksv6 = self._read_blocks()
+        if blocksv4 or blocksv6:
```
caseydavenport, line 514: Might add a comment here explaining why we're preventing this.

calico_containers/pycalico/ipam.py
```diff
@@ -372,16 +490,52 @@ class HostAffinityClaimedError(DataStoreError):
...
-    :return: etcd key as string.
+    :return: etcd key as a string.
     """
     path = IPAM_BLOCK_PATH % {'version': block_cidr.version}
     return path + str(block_cidr).replace("/", "-")
 
 
+def _block_host_key(host, block_cidr):
+    """
+    Translate a block CIDR into the host block key.
```
caseydavenport: into the host block key... used in what?

Should mention that this is the key used to store block affinity in the host etcd tree (or something like that)

calico_containers/pycalico/ipam.py
```diff
@@ -474,11 +629,11 @@ def _auto_assign(self, ip_version, num, handle_id,
                           host, pool)
```
caseydavenport, line 669: https://github.com/projectcalico/libcalico/pull/74/files#diff-17a9750fd8a984de4deef40d0202fa26R600

This should mention the difference in behavior when strict affinity is enabled.

calico_containers/pycalico/ipam.py
```diff
@@ -651,8 +829,10 @@ def assign_ip(self, address, handle_id, attributes, host=None):
                     raise PoolNotFound("%s is not in any configured pool" %
                                        address)
 
-            # Try to assign.  Throws exception if already assigned -- let it.
-            block.assign(address, handle_id, attributes)
+            # Try to assign.  Throws exception if already assigned, or if
```
caseydavenport: What kind of Exception does this throw?

calico_containers/pycalico/ipam.py
```diff
@@ -832,3 +1034,116 @@ def get_assignment_attributes(self, address):
...
+                  [IPNetwork<blocks that were not claimed>],
+                  [IPNetwork<blocks that were claimed by another host>])
+        """
+        assert isinstance(cidr, IPNetwork)
+        if not validate_block_size(cidr):
+            _log.info("Requested CIDR %s is too small", cidr)
+            raise InvalidBlockSizeError("The requested CIDR is smaller than "
+                                        "the minimum block size.")
+        if not host:
+            host = get_hostname()
```
caseydavenport: How about:

```
host = host or get_hostname()
```

calico_containers/pycalico/ipam.py
```diff
@@ -832,3 +1034,116 @@ def get_assignment_attributes(self, address):
...
+        released = []
+        not_claimed = []
+        claimed_by_other = []
+
+        for block_cidr in cidr.subnet(BLOCK_PREFIXLEN[cidr.version]):
+            try:
+                self._release_block_affinity(host, block_cidr)
+            except HostAffinityClaimedByOtherError:
+                claimed_by_other.append(cidr)
+                break
```
caseydavenport: What is the reasoning for wanting to `break` here (and below), rather than continue and release as many of the blocks that we own as we can?

calico_containers/pycalico/ipam.py
```diff
@@ -832,3 +1034,116 @@ def get_assignment_attributes(self, address):
...
+                released.append(block_cidr)
+
+            return released, not_claimed, claimed_by_other
+
+    @handle_errors
+    def release_host_affinities(self, host):
+        """
+        Release host affinities, for all blocks owned by the host.
+
+        :param host: (optional) The host ID to compare against the affinity of
```
caseydavenport, line 1175: Doesn't look optional to me.

calico_containers/pycalico/ipam.py
```diff
@@ -832,3 +1034,116 @@ def get_assignment_attributes(self, address):
...
+        # the block is owned by another host - we simply won't release that
+        # block.
+        _log.debug("Releasing affinities for %s", host)
+        for version in (4, 6):
+            cidrs = self._get_affine_blocks(host, version, None)
+            for cidr in cidrs:
+                try:
+                    self._release_block_affinity(host, cidr)
+                except HostAffinityClaimedByOtherError:
+                    _log.info("Affine block %s is not owned by host %s - skip",
```
caseydavenport, line 1192: Presumably this branch is to handle the race condition where we get the blocks for this host, another client releases one of them, and then we try to release it?

I'd add a comment explaining how we could hit this branch.

calico_containers/pycalico/block.py
robbrockbank: That was just to convert empty string to None.  Not sure why I thought it was necessary.  I'll remove.

caseydavenport:
Ok - first round done.  Some questions and comments for ya.

calico_containers/pycalico/datastore.py
```diff
@@ -501,6 +506,23 @@ def get_ip_pool_config(self, version, cidr):
         return IPPool.from_json(data)
 
     @handle_errors
+    def set_ip_pool_config(self, version, pool):
+        """
+        Set the IP pool configuration.
+
+        :param version: 4 for IPv4, 6 for IPv6
+        :param cidr: IPNetwork object representing the pool
```
caseydavenport: Docstring is wrong.

calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -159,10 +163,13 @@ def to_json(self):
...
         if self.masquerade:
             json_dict["masquerade"] = True
-        # Only write IPAM when set to False, this keeps the behavior the same
-        # for old versions of the data.
+        # Only write "ipam" and "disabled" when they differ from their default
+        # values.  This keeps the interface unchanged between versions when
+        # these fields are not required.
         if not self.ipam:
             json_dict["ipam"] = False
+        if self.disabled:
```
caseydavenport: Is it really a problem if we have a new key in the dictionary? Old clients will just ignore it.

calico_containers/pycalico/ipam.py
```diff
@@ -62,7 +64,7 @@ def _read_block(self, block_cidr):
         key = _block_datastore_key(block_cidr)
         try:
             # Use quorum=True to ensure we don't get stale reads.  Without this
-            # we allow many subtle race conditions, such as creating a block,
+            # we allow many subtle race condiftions, such as creating a block,
```
caseydavenport: Typo: condiftions

calico_containers/pycalico/datastore_datatypes.py
robbrockbank: ta.

calico_containers/pycalico/ipam.py
robbrockbank: Nope - I copied this from another deletion branch which was also wrong.  I tested last night - and is fixed.

calico_containers/pycalico/ipam.py
robbrockbank: You are absolutely correct.  This was a throwback to this method doing a host deletion, which it isn't - it's just removing the affinity.  This branch should go.

calico_containers/pycalico/block.py
```diff
@@ -179,18 +204,20 @@ def auto_assign(self, num, handle_id, attributes, host,
...
+        affinity check is always performed.
         :return: List of assigned addresses.  When the block is at or near
         full, this method may return fewer than requested IPs.
         """
         assert num >= 0
 
+        affinity_check = self.strict_affinity or affinity_check
         if affinity_check and host != self.host_affinity:
-            raise NoHostAffinityWarning("Block host affinity is %s (not %s)" %
+            raise NoHostAffinityError("Block host affinity is %s (not %s)" %
```
caseydavenport, line 220: Indentation is off here

calico_containers/pycalico/ipam.py
```diff
@@ -125,8 +136,36 @@ def _get_affine_blocks(self, host, version, pool):
...
+        """
+        assert isinstance(pool, IPPool)
+
+        # Construct the path
+        hosts_and_blocks = []
+        try:
+            result = self.etcd_client.read(IPAM_HOSTS_PATH, quorum=True, recursive=True).leaves
+            for child in result:
+                packed = child.key.split("/")
+                if len(packed) == 9:
```
caseydavenport, line 160: I don't really understand what is going on here, but it feels kinda yucky.

What is packed?  Why does it have length 9? What does it mean when it doesn't have length 9?

calico_containers/pycalico/datastore_datatypes.py
robbrockbank: k

calico_containers/pycalico/ipam.py
robbrockbank: updated comment

calico_containers/pycalico/ipam.py
robbrockbank: sure - I've updated all occurrences.

calico_containers/pycalico/datastore.py
robbrockbank: dammit

calico_containers/pycalico/ipam.py
caseydavenport, line 160: This at least needs more commenting.

calico_containers/pycalico/ipam.py
robbrockbank: Ah another f creeps in because of my inability to do ctrl-f on a Mac

calico_containers/pycalico/ipam.py
robbrockbank, line 160: This is littered through the whole datastore class and IMHO needs some kind of overhaul to improve code readability.  I'd rather not change this one instance.

caseydavenport, line 160: Okey doke.

caseydavenport:
OK - Code is looking pretty good to me (one or two more comments outstanding I think).  I won't get a chance to look at the tests but someone else should.

calico_containers/pycalico/ipam.py
```diff
@@ -199,13 +242,63 @@ def _claim_block_affinity(self, host, block_cidr):
...
+                          block_cidr, block.host_affinity)
+
+            try:
+                if block.is_empty():
+                    # The block is empty, so just delete the block.
+                    _log.debug("Block is empty - delete block")
+                    self._delete_block(block)
+                else:
+                    # The block is not empty so remove affinity from the block.
+                    # This prevents the host automatically assigning from this
```
TrimBiggs, line 283: This prevents the host **_from**_ automatically...

calico_containers/pycalico/ipam.py
```diff
@@ -199,13 +242,63 @@ def _claim_block_affinity(self, host, block_cidr):
...
+                    # block unless we are allowed to overflow into non-affine
+                    # blocks when affine blocks are exhausted, and provided the
+                    # block is still valid (i.e has a corresponding IP Pool),
```
TrimBiggs: , -> .

calico_containers/pycalico/ipam.py
```diff
@@ -832,3 +1080,136 @@ def get_assignment_attributes(self, address):
...
+        assert isinstance(cidr, IPNetwork)
+        if not validate_block_size(cidr):
+            _log.info("Requested CIDR %s is too small", cidr)
+            raise InvalidBlockSizeError("The requested CIDR is smaller than "
+                                        "the minimum block size.")
+
+        host = host or get_hostname()
+
+        if not self._validate_cidr_in_pools(cidr):
+            _log.info("Requested CIDR %s is not is a configured pool", cidr)
```
TrimBiggs: typo: "is not is"

calico_containers/tests/unit/ipam_test.py
```diff
@@ -1291,6 +1456,42 @@ def m_read(path, quorum):
...
+            return result
+        self.m_etcd_client.read.side_effect = m_read
+
+        ip_pool = IPPool(IPNetwork("10.0.0.0/8"))
+        host_block_pairs = self.client._get_host_block_pairs(ip_pool)
+        assert_list_equal(host_block_pairs, expected_pairs)
+
+    def test_get_host_block_pairs_not_found(self):
+        """
+        Test of _get_host_block_pairs() when host key is not found,
```
TrimBiggs: , -> .

TrimBiggs:
Some minor comments about some grammar, but otherwise LGTM.

PR #81, opened 2016-01-25
Build calico/build from local libcalico code

tomdee:
Docker hub is set up so that when we tag a release we'll automatically get a tagged docker image


caseydavenport:
LGTM.

PR #84, opened 2016-02-06
Add contributing MD

robbrockbank:


robbrockbank:
Will rework based on calico_containers one (once that one is approved)

PR #91, opened 2016-03-04
Catch malformed etcd_authority variables and raise a datastore error

### LINKED ISSUE #85: libcalico yields misleading error from bad ETCD_AUTHORITY env var
If `ETCD_AUTHORITY` environment is missing a colon, pycalico will fail to initialize the `IPAMClient` with  a misleading stack trace. 

```
$ ETCD_AUTHORITY=erroneous python -c "import pycalico.ipam; pycalico.ipam.IPAMClient()"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pycalico/datastore.py", line 137, in __init__
    (host, port) = etcd_authority.split(":", 1)
ValueError: need more than 1 value to unpack
```

[Link to line 137](https://github.com/projectcalico/libcalico/blob/master/calico_containers/pycalico/datastore.py#L137)

I propose that we should catch the value error and reword it to state "malformed ETCD_AUTHORITY", possibly in the form of a new MalformedEtcdAuthority exception?

ozdanborne:
Fixes #85

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -25,7 +25,7 @@ from pycalico.datastore_datatypes import Rules, BGPPeer, IPPool, \
     Endpoint, Profile, Rule, IF_PREFIX
 from pycalico.datastore_errors import DataStoreError, \
     ProfileNotInEndpoint, ProfileAlreadyInEndpoint, MultipleEndpointsMatch
-from pycalico.util import get_hostname
+from pycalico.util import get_hostname, validate_hostname_port
 
 ETCD_AUTHORITY_DEFAULT = "127.0.0.1:2379"
 ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
@@ -134,8 +134,12 @@ class DatastoreClient(object):
 
     def __init__(self):
         etcd_authority = os.getenv(ETCD_AUTHORITY_ENV, ETCD_AUTHORITY_DEFAULT)
-        (host, port) = etcd_authority.split(":", 1)
+        if not validate_hostname_port(etcd_authority):
+            raise DataStoreError("Invalid %s. It must take the form "
+                                 "<address>:<port>. Value provided is '%s'" %
+                                 (ETCD_AUTHORITY_ENV, etcd_authority))
 
+        (host, port) = etcd_authority.split(":", 1)
         etcd_scheme = os.getenv(ETCD_SCHEME_ENV, ETCD_SCHEME_DEFAULT)
         etcd_key = os.getenv(ETCD_KEY_FILE_ENV, '')
         etcd_cert = os.getenv(ETCD_CERT_FILE_ENV, '')
--- a/calico_containers/pycalico/util.py
+++ b/calico_containers/pycalico/util.py
@@ -2,8 +2,12 @@ import socket
 import sys
 import os
 import re
+import logging
 from netaddr import IPNetwork
 from subprocess import check_output, CalledProcessError
+
+_log = logging.getLogger(__name__)
+_log.addHandler(logging.NullHandler())
 
 HOSTNAME_ENV = "HOSTNAME"
 
@@ -159,21 +163,30 @@ def validate_hostname_port(hostname_port):
     :return: Boolean: True if valid, False if invalid
     """
     # Should contain a single ":" separating hostname and port
-    hostname_port = hostname_port.split(":")
-    if len(hostname_port) != 2:
+    if not isinstance(hostname_port, str):
+        _log.error("Must provide string for hostname:port validation, not: %s" % type(hostname_port))
+        return False
+
+    try:
+        (hostname, port) = hostname_port.split(":")
+    except ValueError:
+        _log.error("Must provide a string splittable by ':' for hostname-port.")
         return False
 
     # Check the hostname format.
-    if not validate_hostname(hostname_port[0]):
+    if not validate_hostname(hostname):
         return False
 
     # Check port range.
     try:
-        port = int(hostname_port[1])
+        port = int(port)
     except ValueError:
+        _log.error("Port must be an integer.")
         return False
-    else:
-        return 1 <= port <= 65535
+    if port < 1 or port > 65535:
+        _log.error("Provided port (%d) must be between 1 and 65535." % port)
+        return False
+    return True
 
 def validate_hostname(hostname):
     """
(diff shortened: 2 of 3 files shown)


TrimBiggs:
LGTM

TrimBiggs:
We should have a test for this.

calico_containers/pycalico/datastore.py
```diff
@@ -134,7 +134,15 @@ class DatastoreClient(object):
 
     def __init__(self):
         etcd_authority = os.getenv(ETCD_AUTHORITY_ENV, ETCD_AUTHORITY_DEFAULT)
-        (host, port) = etcd_authority.split(":", 1)
+        try:
+            (host, port) = etcd_authority.split(":", 1)
```
TrimBiggs: I don't think this will catch if the authority is ":1234", but I think it will be pretty obvious when it fails elsewhere..

TrimBiggs:
LGTM

calico_containers/pycalico/datastore.py
caseydavenport: I think the most robust way to do this is to define a regex pattern which matches valid etcd authorities and parse it using that.

For example, this won't catch if someone puts in `http://host:port` as their etcd authorit (which is a common and easy mistake for users to make)

caseydavenport:
As I said in a comment, I think this should be using regex to do more comprehensive checking on the provided value.

calico_containers/pycalico/datastore.py
caseydavenport: Rather, it will likely catch the error below but the error message won't be great.

e.g

```
ERROR: Invalid ETCD_AUTHORITY port: 'host:port'
```

Rather than:

```
ERROR: ETCD_AUTHORITY 'http://host:port' should not include scheme 'http'
```

Alternatively, we might just want to handle that case gracefully...

TrimBiggs: Good point.  Just need to keep in mind the valid characters for a hostname if using hostname vs ip.

ozdanborne:
Maybe. We do, however, accept both hostnames and IPs. Therefore, our regex statement would  effectively be searching for two sets of values split by a colon, which is the same as what this code does.

TrimBiggs:
There could be a utils `validate_etcd_authority` function:
Some regex check like `((\d{1,3}\.){3}(\d{1,3})|([\s\d]{8,32}):\d{1,5})` that checks both IP and hostname (hostname check there could be better or made need to be different).

This would be useful for when we allow a list of etcd authorities also.  We might want check for an optional `(http://|https://|etcd://){0,1}` if we start allowing this, but we could have wrappers around this function that get the IP/port regardless of what style of etcd authority is passed in.

ozdanborne:
I realized that we already do `ETCD_AUTHORITY` validations in calico-containers utils.py, so I am changing this PR to copy those validations into libcalico for use by other orchestrators (in this case, libcalico itself). 

I'm going to try to keep the scope of this small so it is just a 1:1 move - the only change in this PR besides the move is calling the function from datastore.py ([see second commit](https://github.com/djosborne/libcalico/commit/8a01ecfaf0a6c619fcc99a27605b1e064c4a8650))

@TrimBiggs @caseydavenport I believe this takes care of all your comments (except multiple etcd_authorities and http/s, but I'd prefer to save those for a later PR). Sound good?

Additionally, I will open up a seperate PR to remove this functionality from calico_containers once this one is merged

calico_containers/pycalico/datastore.py
```diff
@@ -134,7 +134,12 @@ class DatastoreClient(object):
 
     def __init__(self):
         etcd_authority = os.getenv(ETCD_AUTHORITY_ENV, ETCD_AUTHORITY_DEFAULT)
-        (host, port) = etcd_authority.split(":", 1)
+        if not validate_hostname_port(etcd_authority):
+            raise DataStoreError("Invalid %s. It must take the form <address>:<port>. "
+                            "Value provided is '%s'" % (ETCD_AUTHORITY_ENV,
```
TrimBiggs: Alignment is off here

TrimBiggs:
I'm ok with that.

This LGTM with a minor comment.  Are you planning to have a PR that removes this from calico-containers?

calico_containers/pycalico/util.py
```diff
@@ -150,3 +150,44 @@ def validate_icmp_type(icmp_type):
...
+def validate_hostname_port(hostname_port):
+    """
+    Validate the hostname and port format.  (<HOSTNAME>:<PORT>)
+    An IPv4 address is a valid hostname.
+
+    :return: Boolean: True if valid, False if invalid
+    """
+    # Should contain a single ":" separating hostname and port
+    hostname_port = hostname_port.split(":")
+    if len(hostname_port) != 2:
```
caseydavenport: Might be nice to have some debug logging in this function so we can track where validation went sour.

ozdanborne: Might be nice to have logging in libcalico :frowning: Afaik we don't currently, right?

calico_containers/pycalico/util.py
```diff
@@ -150,3 +150,44 @@ def validate_icmp_type(icmp_type):
...
+
+def validate_hostname_port(hostname_port):
+    """
+    Validate the hostname and port format.  (<HOSTNAME>:<PORT>)
+    An IPv4 address is a valid hostname.
+
+    :return: Boolean: True if valid, False if invalid
+    """
+    # Should contain a single ":" separating hostname and port
+    hostname_port = hostname_port.split(":")
```
caseydavenport: Any possibility of something that is not a string being passed here?  Do we need to assert that `hostname_port` is a string before we try to split it?

calico_containers/pycalico/util.py
caseydavenport: We do have logging in libcalico (though it appears we don't in util.py).

See netns.py, ipam.py, handle.py, block.py

It is up to the calling code to display those logs or not.

caseydavenport:
Looks fine to me.

I'd like to see the corresponding PR in `calico-containers` which removes this function.

calico_containers/pycalico/util.py
ozdanborne: I wouldn't mind adding logging to it. This PR or a new one?

ozdanborne:
@caseydavenport I haven't opened the PR as it will just fail CI tests: https://github.com/projectcalico/calico-containers/compare/master...djosborne:remove-validate-hostnames?expand=1

calico_containers/pycalico/datastore.py
```diff
@@ -134,7 +134,12 @@ class DatastoreClient(object):
 
     def __init__(self):
         etcd_authority = os.getenv(ETCD_AUTHORITY_ENV, ETCD_AUTHORITY_DEFAULT)
-        (host, port) = etcd_authority.split(":", 1)
+        if not validate_hostname_port(etcd_authority):
+            raise DataStoreError("Invalid %s. It must take the form "
+                                 "<address>:<port>. Value provided is '%s'" %
+                                 (ETCD_AUTHORITY_ENV, etcd_authority))
+        else:
```
caseydavenport: You don't need the `else` here - happy to let it go for now but is cleaner to not include it.

calico_containers/pycalico/util.py
```diff
@@ -150,3 +154,61 @@ def validate_icmp_type(icmp_type):
...
+    Validate the hostname and port format.  (<HOSTNAME>:<PORT>)
+    An IPv4 address is a valid hostname.
+
+    :return: Boolean: True if valid, False if invalid
+    """
+    # Should contain a single ":" separating hostname and port
+    if not isinstance(hostname_port, str):
+        _log.error("Must provide string for hostname:port validation, not: %s" % type(hostname_port))
+        return False
+    try:
```
caseydavenport, line 170: Should have a newline before this try.

caseydavenport:
Two minor comments - fix them if it pleases you.

Then, squash commits, and then LGTM.

calico_containers/pycalico/util.py
ozdanborne, line 170: Aesthetically, I'd agree that it looks better, and I'll change it. I'm curious if there's a pep8 that also says there should be a new line here? And what's the gist of it? Always add a new line after conclusion of an if statement?

PR #95, opened 2016-03-11
Add get_hosts_data function to get bgp node info for all nodes

TrimBiggs:
This function is used by the new `calicoctl node show` command, which outputs hostname, IP, and peer information about the nodes in the cluster.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -76,8 +76,10 @@ BGP_HOST_AS_PATH = BGP_HOST_PATH + "as_num"
 BGP_HOST_PEERS_PATH = BGP_HOST_PATH + "peer_v%(version)s/"
 BGP_HOST_PEER_PATH = BGP_HOST_PATH + "peer_v%(version)s/%(peer_ip)s"
 
-# Grabs hostname from etcd datastore key
-HOSTNAME_DATASTORE_RE = re.compile(BGP_HOSTS_PATH + "(.*)/ip_addr_v[46]")
+# Grabs hostname from etcd datastore keys
+HOSTNAME_IP_DATASTORE_RE = re.compile(BGP_HOSTS_PATH + "(.*)/ip_addr_v[46]")
+HOSTNAME_ANY_BGP_DATASTORE_RE = re.compile(BGP_HOSTS_PATH +"(.*)/" +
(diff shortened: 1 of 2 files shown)


calico_containers/pycalico/datastore.py
```diff
@@ -76,8 +76,10 @@
 BGP_HOST_PEERS_PATH = BGP_HOST_PATH + "peer_v%(version)s/"
 BGP_HOST_PEER_PATH = BGP_HOST_PATH + "peer_v%(version)s/%(peer_ip)s"
 
-# Grabs hostname from etcd datastore key
-HOSTNAME_DATASTORE_RE = re.compile(BGP_HOSTS_PATH + "(.*)/ip_addr_v[46]")
+# Grabs hostname from etcd datastore keys
+HOSTNAME_IP_ETCD_RE = re.compile(BGP_HOSTS_PATH + "(.*)/ip_addr_v[46]")
```
tomdee: The word datastore is preferred to etcd. What was your reasoning behind changing it?

calico_containers/pycalico/datastore.py
```diff
@@ -365,6 +367,47 @@ def remove_host(self, hostname):
...
+            # Get all host data
+            host_data = self.etcd_client.read(BGP_HOSTS_PATH, recursive=True)
+        except EtcdKeyNotFound:
+            # No BGP hosts currently configured in etcd
+            return {}
+
+        host_dict = {}
+        for host_leaf in host_data.leaves:
+            # Match expected host data values
+            match = HOSTNAME_ANY_BGP_ETCD_RE.match(host_leaf.key)
```
tomdee: I would combine this line with the next

calico_containers/pycalico/datastore.py
TrimBiggs: I felt that ETCD was easier to read/shorter. It seems like there is plenty of use for each of the terms in this file and there is only one regex global (just used here) so I figured I'd change it.

If we're preferring datastore over etcd, I can change it back and rename the below global as well.

tomdee:
Looks good to me

TrimBiggs:
@tomdee Mind giving the tests a review?

tomdee:
Yep, LGTM

PR #100, opened 2016-03-23
Add contributing doc to repo

TrimBiggs:


CONTRIBUTING.md
```diff
@@ -0,0 +1,143 @@
...
+   commands that were run)
+-  The expected result and actual result.
+-  Versions of appropriate binaries and libraries.  For example, the output from
+   `calicoctl version`, your version of Docker, rkt, Kubernetes etc.
+-  A link to any diagnostics (e.g. if using `calicoctl`, you can gathered
+   diagnostics using `calicoctl diags` - this provides instructions for 
+   uploading the diags bundle to transfer.sh - or alternatively if the 
+   diagnostics contains sensitive information we can set up an alternative
+   method for transfer
+-  If using `calicoctl` the output from `calicoctl status` run on each node
```
TrimBiggs: I'm not sure these are relevant for libcalico since it is likely the API that someone is using, not calicoctl.

robbrockbank: no, most of this is not particularly relevant.  just include things that probably are:  version or commit ID of library,  stack trace etc.

robbrockbank: Indeed.  I think this section will need a rework.  I think you can assume people are using this as a Python library rather than part of a calico-containers deployment.

robbrockbank:
lgtm

PR #101, opened 2016-03-25
Remove the concept of next_hops/gateway from the libcalico code

tomdee:
It's just not needed


calico_containers/pycalico/netns.py
```diff
@@ -182,23 +182,29 @@ def set_veth_mac(veth_name_host, mac):
...
     :param veth_name_ns: The name of the interface in the namespace.
     :return: None. Raises CalledProcessError on error.
     """
-    assert isinstance(next_hop, IPAddress)
     with NamedNamespace(namespace) as ns:
+        # Use the first link local address as the dummy address.
+        # This only ever appears in an ARP packet.
+        # The host has proxy arp configured on the other end of the veth so
+        # will respond with the MAC of the veth
+        dummy_next_hop_4 = "169.254.0.0"
```
matthewdupre: This one isn't really a link local address: while 169.254/16 is generally used for link local addresses, the first and last 256 addresses are somehow special.  See https://tools.ietf.org/html/rfc3927 section 8.

.1.0 seems safer.

matthewdupre:
This all looks fine.  The only slight concern I had was that the first IPv6 address on the host side of the veth might not be IPv6 link local; but that suggests someone/something is meddling with us (we don't put any IP there) so much that we don't need to worry about it.

PR #102, opened 2016-04-04
Randomise selection order for affine blocks.

fasaxc:
I think this is what's needed to avoid contention when many hosts assign IPAM blocks concurrently.  I'm just about to try it in a scale test, though...

### DIFF
--- a/calico_containers/pycalico/ipam.py
+++ b/calico_containers/pycalico/ipam.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections import deque
 
 from etcd import EtcdKeyNotFound, EtcdAlreadyExist, EtcdCompareFailed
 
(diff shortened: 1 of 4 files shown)


robbrockbank:
Are you sure you want this to be random?  If you have multiple clients on the same host attempting to allocate IP addresses for the first time you'll end up with multiple allocation blocks being assigned on the host.

Wouldn't it be better to use a hash of the host to provide a fix entry point into the CIDRs for a specific host?

fasaxc:
Hadn't thought of that, good spot.  I'll make that change and put it through its paces in my scale testing.

fasaxc:
Seems to do the trick after I finally managed to get a working container image.  I did spot that there's some pretty similar function in `IpamClient.random_blocks()` but my version avoids building up a huge list of blocks and shuffling them.  Occurred to me that, for IPv6, enumerating the blocks might be a little time-consuming ;-)  WDYT?

robbrockbank:
Yeah I've always worried that the shuffling code will be the consuming.  Wanna fix that up too?  :-)

fasaxc:
@robbrockbank OK, I've made that change.  Means I can now use `self._random_blocks` in `_new_affine_blocks` too, which makes it DRYer.

calico_containers/pycalico/ipam.py
```diff
@@ -649,25 +636,91 @@ def _auto_assign(self, ip_version, num, handle_id,
...
+                    allocated_ips.extend(ips_from_random_blocks)
+        _log.info("Allocated %s of %s requested IPs", len(allocated_ips), num)
+        return allocated_ips
+
+    def _allocate_ips_explicit_blocks(self, blocks, num, attributes, handle_id,
+                                      host):
+        """Tries to allocate IPs from the explicitly-listed blocks.
+
+        :param list blocks: Blocks to allocate from (for example, the affine
+               blocks for a host).
```
robbrockbank: nit: formatting off on this line

calico_containers/pycalico/ipam.py
```diff
@@ -1241,3 +1313,80 @@ def remove_ipam_host(self, host):
...
+
+    base_subnet_addr = str(cidr.cidr.ip)  # Throws away the .1 in 10.0.0.1/8.
+    num_returned = 0
+    # Choose our step and initial position randomly.  We avoid using
+    # rnd.shuffle() because that would require us to generate the whole list
+    # of CIDRs, and there could be millions of those!
+    step = rnd.choice(STEPS)
+    position = rnd.randint(0, max_subnets - 1)
+    while num_returned < max_subnets:
+        subnet = cidr.__class__('%s/%d' % (base_subnet_addr, prefixlen),
```
robbrockbank: cidr can only be an IPNetwork no?  If so, why not just use IPNetwork()?

robbrockbank:
Looks good to me.  Couple of minor comments.  Took me a while to get my head round the random blocks thing... not sure if because I'm slow, or whether it just needs some beefing up of the comments.

fasaxc:
Added more comments, and the license for netaddr.  Bit of a shame to have to pick up more license than code but I think that's necessary.

robbrockbank:
lol - wish you'd put that comment in there sooner - I had to get my maths head on... it's been a long time ;-)

Looks good.  Need to resolve conflicts though, then we're good to go.

PR #105, opened 2016-04-22
Change get_host_ips to return IPAddress objs, not strings

TrimBiggs:


robbrockbank:
This is an interface change and we do have a number of components using libcalico.  Did you check if other components make use of it?

TrimBiggs:
I've just gone through each repo I had not previously checked, only libcalico and calico-containers use this util function.

robbrockbank:
LGTM

PR #109, opened 2016-05-06
Disable felix status reporting by default.

fasaxc:
Fixes https://github.com/projectcalico/calico-containers/issues/970

Felix supports reporting its status into etcd, which we use in the OpenStack integration to report the agent status. That feature is currently turned on in the container integration even though we're not exposing it anywhere.

At high scale, it uses a significant chunk of etcd write bandwidth (the default is to write 2 etcd keys every 30s from every felix).

We should turn this off until it's needed.


caseydavenport:
LGTM

PR #110, opened 2016-05-06
Add python files as a Makefile dependency for the build image.

fasaxc:
Ensures changes to python files trigger rebuild of the image.


tomdee:
Good spot. LGTM, merging

PR #111, opened 2016-05-06
Add and test new validate_port_spec utility

nelljerram:
validate_port_spec validates the port specifications that can be
specified on the calicoctl command line.  This is different from
validate_ports, which validates the port values that can occur in the
etcd data model.


matthewdupre:
See comment in [#972](https://github.com/projectcalico/calico-containers/pull/972#issuecomment-218535471).

nelljerram:
Thanks @matthewdupre, PR updated.  Please could you rereview and merge?

PR #117, opened 2016-06-08
Changing util's validation to raise on exceptions

insequent:
These validation functions are used by clients such as calico-containers, and when they fail, users may want to know why such validations are failing. With this feature, pre-existing code will continue to work as is, however clients now have the option of returning the error (exception) for further diagnostics.

It feels a touch clunky with the way this is implemented, but I didn't want to break any existing code, and this seemed the simplest way. I'm definitely open to any ideas on alternatives.

### DIFF
--- a/calico_containers/pycalico/util.py
+++ b/calico_containers/pycalico/util.py
@@ -1,15 +1,19 @@
-from netaddr.core import AddrFormatError
+#!/usr/bin/python
+"""General utility functions"""
 
-import netaddr
 import socket
 import sys
 import os
 import re
 import logging
-from netaddr import IPNetwork, IPAddress
 from subprocess import check_output, CalledProcessError
 
-_log = logging.getLogger(__name__)
+import netaddr
+from netaddr import IPNetwork, IPAddress
+from netaddr.core import AddrFormatError
+
+
+_log = logging.getLogger(__name__)  # pylint: disable=invalid-name
 _log.addHandler(logging.NullHandler())
 
 HOSTNAME_ENV = "HOSTNAME"
@@ -28,9 +32,60 @@ IPV4_RE = re.compile(r'inet ((?:\d+\.){3}\d+)/\d+')
 IPV6_RE = re.compile(r'inet6 ([a-fA-F\d:]+)/\d{1,3}')
 
 
+class ValidationError(ValueError):
+    """Base class for all validation errors"""
+    pass
+
+
+class AddrValidationError(ValidationError, AddrFormatError):
+    """Error when passed value cannot be converted to an IP Address"""
+    pass
+
+
+class CharValidationError(ValidationError):
+    """
+    Error when passed string includes incompatible characters or is
+    missing required ones.
+    """
+    pass
+
+
+class RangeValidationError(ValidationError):
+    """Error when passed value is outside the range of valid values"""
+    pass
+
+
+class TypeValidationError(ValidationError):
+    """Error when passed value is an invalid type"""
+    pass
+
+
+class VersionMismatchError(ValidationError):
+    """Error when passed IP Version does not match the CIDR/IP"""
+    pass
+
+
+def _return_bool(func, *args, **kwargs):
+    """
+    Simple function to catch exceptions and return a Bool
+
+    :param func: The function being run
+    :param *args: All positional args being passed to func
+    :param **kwargs: All key-value args being passed to func
+    :return: True if function succeeds, False if exception is raised
+    :rtype: bool
+    """
+    try:
+        func(*args, **kwargs)
+    except ValidationError:
+        return False
+    else:
+        return True
+
+
 def generate_cali_interface_name(prefix, ep_id):
-    """Helper method to generate a name for a calico veth, given the endpoint
-    ID
+    """
+    Helper method to generate a name for a calico veth, given the endpoint ID
 
     This takes a prefix, and then truncates the EP ID.
 
@@ -67,7 +122,7 @@ def get_host_ips(version=4, exclude=None):
     try:
         ip_addr_output = check_output(["ip", "-%d" % version, "addr"])
     except (CalledProcessError, OSError):
-        print "Call to 'ip addr' Failed"
+        print("Call to 'ip addr' Failed")
         sys.exit(1)
 
     # Separate interface blocks from ip addr output and iterate.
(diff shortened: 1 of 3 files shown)


calico_containers/pycalico/util.py
```diff
@@ -290,25 +351,55 @@ def validate_cidr_versions(cidrs, ip_version=None):
...
-        return False
-    return True
+        err = ValueError("CIDR is invalid. " + str(exc).capitalize())
+        return err if return_error else False
+
+    return None if return_error else True
 
 
-def validate_ip(ip_addr, version):
+def validate_ip(ip_addr, version=None, return_error=False):
```
insequent: Just remembered that I added `version=None` but didn't do anything with it. This validation function is so different than the rest, that I was hesitant  to touch it (none of the others have an assert). For simplicity sake, I'll probably just take out the `=None` addition, but there's really no reason to require a version when validating an IP unless you are explicitly checking for the version, which the name of this method doesn't imply.

insequent:
Added logic for `version=None` in `validate_ip()`. Seems more useful for it to be able to handle no version and just check if the IP is a valid.

Also, fixed my email to use the one I signed the CLA with.

insequent:
Last update pending review (I promise). I noticed multiple functions were missing complete docstrings for the new return_error param, and one had no param or return at all. Added those in. This was a purely cosmetic update.

matthewdupre:
Thanks for contributing @insequent!  I'm sorry we got this wrong from the start.  I'll make any detailed comments against the code, but first cover the general form of the validation functions.

I had a chat with some of the team, and we think the least clunky (but still back compatible) approach is probably to rename the functions to verify_, change them to return nothing and raise exceptions when needed, and then add the validate_ functions back in to point to them.

Maybe we could decorate the old validate_ functions to log a deprecation warning, call the new verify_ equivalent and return accordingly?

What do you think?

calico_containers/pycalico/util.py
```diff
@@ -135,152 +144,211 @@ def validate_ports(port_list):
...
+
     try:
         valid = 0 <= int(icmp_type) < 255
     except ValueError:
         valid = False
-    return valid
 
+    if not valid and return_error:
+        err = ValueError("ICMP type is invalid. Value must be between 0 and "
+                         "255 ('{0}' given).".format(icmp_type))
```
matthewdupre: The range includes 0 (but not 255).

calico_containers/pycalico/util.py
```diff
@@ -135,152 +144,211 @@ def validate_ports(port_list):
...
-        return False
+        err_mess = "Hostname must be a string, not {0}".format(hostname)
+        _log.error(err_mess)
+        return TypeError(err_mess) if return_error else False
     hostname_len = len(hostname)
     if hostname_len > 255:
-        _log.error("Hostname length (%d) should be less than 255 characters."
-                   % hostname_len)
-        return False
+        err_mess = ("Hostname length (%d) should be less than 255 characters."
```
matthewdupre: less than 256

matthewdupre:
Detail-wise, I found a couple of off-by-ones in the messages, but otherwise looks great - thanks :).

insequent:
That makes sense and sounds much cleaner. I'll get right on it!

djosborne:
@matthewdupre @insequent The suggested change is close, but not backwards compatible. We should not rename the functions, but simply create new ones which raise the exceptions.

insequent:
This commit should be closer in line with what you're looking for. However, I did take a few liberties this time:
1. I changed the order of the functions to be alphabetical (easier to find that way)
2. ICMP Types are now inclusively 0 to 255. I have no qualms changing this back, but it seems some people do use 255 as ICMP type (iptables in particular).
3. I changed everything I could to satisfy pylint/pep8 (cosmetic changes). The too long lines for `MOCK_IP_ADDR` and its friends in the test_util.py could not be changed without making the string look ugly, so I left those as is (you could do some trickiness with string concatenation, but the full example is much easier to read altogether).

insequent:
Oh, as to deprecated decorators, I love the idea, but wasn't sure how you all would want them. As you can see in the code, I just put `DEPRECATED` into the docstrings. Preferably, you would make a custom decorator that logs a warning, however the logging in these modules are super limited (and uses a NullHandler?).

calico_containers/pycalico/util.py
```diff
@@ -104,211 +108,434 @@ def get_hostname():
...
-            in_range = (len(ports) == 2) and (int(ports[0]) < int(ports[1])) \
-                       and validate_ports(ports)
-        else:
-            try:
-                in_range = 0 <= int(port) < 65536
-            except ValueError:
-                in_range = False
-        if not in_range:
-            break
+    return return_bool(validate_asn, asn)
```
matthewdupre: Should call verify_, not validate_.

calico_containers/pycalico/util.py
```diff
@@ -104,211 +108,434 @@ def get_hostname():
...
+    DEPRECATED (use verify_port_str)
+
+    Checks whether the command line word specifying a set of ports is valid.
+
+    :param port_str: A comma delimited list of ports and port ranges
+    :type port_str: str
+    :return: returns True if ports are in range, False if not in range
+    :rtype: bool
+    """
+    return validate_ports(port_str.split(","))
```
matthewdupre: Missing return_bool() I think?

calico_containers/pycalico/util.py
insequent: Oh! I didn't even notice there's no unit tests for `validate_asn`. I'll fix that as well.

insequent:
Added unit tests for `validate_asn` and `verify_asn`.  These will catch typos like the one I made, swapping the function names on accident. Note that I had to rework the try/except block for `verify_asn`, because the outer try block was catching all the raises inside the try block (another bug found by the unit tests).

calico_containers/pycalico/util.py
```diff
@@ -104,211 +108,450 @@ def get_hostname():
...
+    return return_bool(verify_cidr, cidr)
+
+
+def verify_cidr(cidr):
+    """
+    Validate cidr is in correct CIDR notation
+
+    :param cidr: IP addr and associated routing prefix
+    :type cidr: str
+    :return: True if valid IP, False if invalid
```
matthewdupre: Should be returns None.

calico_containers/pycalico/util.py
```diff
@@ -1,16 +1,20 @@
...
 from subprocess import check_output, CalledProcessError
 
-_log = logging.getLogger(__name__)
-_log.addHandler(logging.NullHandler())
+import netaddr
+from netaddr import IPNetwork, IPAddress
+from netaddr.core import AddrFormatError
+
+
+LOG = logging.getLogger(__name__)
```
matthewdupre: I know LOG is quite widely used, but the rest of the code uses _log, and I'd prefer to keep it internally consistent - can you please revert this?

calico_containers/pycalico/util.py
```diff
@@ -104,211 +108,450 @@ def get_hostname():
         return socket.gethostname()
 
 
-def validate_port_str(port_str):
+def return_bool(func, *args, **kwargs):
```
matthewdupre: I think _return_bool() is better here: it's not intended to be part of the API.

calico_containers/pycalico/util.py
```diff
@@ -104,211 +108,450 @@ def get_hostname():
...
+    :param func: The function being run
+    :param *args: All positional args being passed to func
+    :param **kwargs: All key-value args being passed to func
+    :return: True if function succeeds, False if exception is raised
+    :rtype: bool
     """
-    return validate_ports(port_str.split(","))
+    try:
+        func(*args, **kwargs)
+    except Exception:  # pylint: disable=broad-except
```
matthewdupre: I'd like us to use a custom Exception class for this (VerifyError seems like the obvious choice).  Motivation being that it propagates unforeseen Exceptions up the stack; clearly distinguishing verification failures (user error) from coding errors.  It also reduces risk that someone writing against this module misses an exception.  We do lose a little expressiveness (currently TypeError vs ValueError), but I think for user input validation the thing that really matters is the message.

matthewdupre:
OK - I promise this is the last round of comments!  Thanks for a very comprehensive PR!

insequent:
Whew! Ok, so I took things even further, so definitely double check me.
1. Made specific custom exceptions
2. Changed `(verify|validate)_hostname` to allow IPv6 hostnames. This is kinda a big one, so definitely let me know if you want me to change that back.
3. Changed `LOG` to `_log`. I may make a PR for the whole module at some point, unless you all just really like the current name (technically a pep8 violation).
4. Changed `return_bool` to `_return_bool`.
5. Changed try blocks to be as specific as possible. This is what I originally thought you were talking about @matthewdupre, not just `_return_bool` (oops).

calico_containers/pycalico/util.py
insequent: This comment is invalid now, although Github's page doesn't seem to think so. Thought it was my browser caching it, but it shows the same way in others. _shrug_

matthewdupre: I guess the comment might have moved from verify_ to validate_ as the diff changed underneath it?  Either that or it was just a bogus comment in the first place.

Weird, but you're right - it's definitely fixed.

calico_containers/pycalico/util.py
```diff
@@ -104,211 +150,505 @@ def get_hostname():
...
-        return False
+
+    if not 0 < hostname_len < 255:
+        err_mess = ("Hostname length can only be 1 to 254 chars long (length "
+                    "{0} given)".format(hostname_len))
+        _log.error(err_mess)
+        raise RangeValidationError(err_mess)
+
+    # NOTE: The real limit in DNS is 255 octets (253 chars) or 254 chars
+    #       if you include the root domain (i.e. a period on the end). RFC1035
```
matthewdupre, line 388: Nice.

matthewdupre:
1. One remaining thing - I'd like a single custom base class, so people can just catch everything, regardless of what validation we add under the covers.
2. I'm OK with taking this, although I'm not 100% on how it would work in practice (1::2 vs 1::0:2 etc.), 
3. You're right, but I'm not sure it's worth making the change.  I'll ask the team about pointing a linter at PRs though (maybe https://github.com/markstory/lint-review or something?).
4. Sorry if I wasn't clear!

insequent:
That makes perfect sense. I'll add single base class.

insequent:
The catch-all exception class is now `ValidationError`.

matthewdupre:
Sorry it took me so long to get back to this: all looks great now.

Thanks for the contribution :)

PR #118, opened 2016-06-10
Change GetEndpoint interface

tomdee:


caseydavenport:
LGTM

lib/endpoint.go
```diff
@@ -139,14 +139,14 @@ func (e *Endpoint) Write(etcd client.KeysAPI) error {
 	return nil
 }
 
-func GetEndpoint(etcd client.KeysAPI, w Workload) (bool, Endpoint, error) {
+func GetEndpoint(etcd client.KeysAPI, w Workload) (*Endpoint, error) {
```
caseydavenport, line 142: Is your long-term plan that the libcalico user will need to pass a KeysAPI on each API call?

Seems like we should be abstracting that from the user of libcalico.

lib/endpoint.go
```diff
@@ -20,7 +20,7 @@ type Endpoint struct {
 	State          string `json:"state"`
 	Name           string `json:"name"`
 	Mac            string `json:"mac"`
-	ProfileID      []string `json:"profile_ids"`
+	ProfileIDs     []string `json:"profile_ids"`
 	IPv4Nets       []string `json:"ipv4_nets"`
 	IPv6Nets       []string `json:"ipv6_nets"`
 	Labels         map[string]string `json:"labels,omitempty"`
```
caseydavenport, line 26: Formatting is a bit off -  I believe the `json:""` bits should be aligned.  Probably just need to run gofmt?

lib/endpoint.go
caseydavenport, line 142: @tomdee is this right?

PR #120, opened 2016-06-14
Update slack URL

eepyaich:


ozdanborne:
Sad that `https://` is broken for `slack.projectcalico.org` :( 

```
Certificate Error
There are issues with the site's certificate chain (net::ERR_CERT_COMMON_NAME_INVALID).
```

@eepyaich have you spoken to @liljenstolpe about that to see if we can fix it?

eepyaich:
Mmm - I did think that was a little sad, but also thought life was too short to try to fix it.  I'll check with "the IT man" though...

eepyaich:
Hold off on this while we try to fix the HTTPS stuff.

liljenstolpe:
We don't actually NEED to fix this.  The slackin process only collects the e-mail.  If it directs them to login, that is directed at the https url.  Everything from the slackin process back is SSL/TLS.  It's just the slackin front-end that's not SSL.  Given that it costs either $20/month or $35/month right now on heroku for an SSL endpoint, not sure  it's worth it.  If we all agree, I'll turn off the SSL on Heroku, and go back to the way it was earlier last week.

There is a upcoming free SSL option on heroku, and I'll will try and get a beta invite for that.

eepyaich:
Chris has now got HTTPS working, so updating this PR.

PR #121, opened 2016-06-14
Add curl into the test container

robbrockbank:
It's used for resetting etc.

Looks like this has been a problem for a while, but the test container had not been rebuilt so everything was still working.  Looks like a push to this repo resulted in a rebuild of the test container.


ozdanborne:
Ah figured it out. Our past builds of libcalico/test include this message:

```
World updated, but the following packages are not removed due to:
  curl: git
```

Looks like in the past, the build process refused to uninstall curl. I believe 'git' no longer requires 'curl, which means our call to `apk remove curl` now succeeded.

LGMT

PR #124, opened 2016-06-24
Generic resource processing [WIP]

robbrockbank:
Initial review of generic resource parsing and validation.


lib/backend/client/client.go
```diff
@@ -0,0 +1,77 @@
...
+		return err
+	}
+	keys := etcd.NewKeysAPI(client)
+	c.etcdClient = &client
+	c.etcdKeysAPI = &keys
+	c.connected = true
+	return nil
+}
+
+func (c *Client) Tiers(namespace string) TierInterface {
```
fasaxc: What's a namespace?  That's not a Calico concept, is it?

lib/backend/client/hostendpoint.go
```diff
@@ -0,0 +1,52 @@
...
+type hostEndpoints struct {
+	r  *Client
+}
+
+// newServices returns a services
+func newHostEndpoints(c *Client) *hostEndpoints {
+	return &hostEndpoints{c}
+}
+
+// List takes a Metadata, and returns the list of hot endpoints that match that Metadata
```
fasaxc: Typo: hot

lib/backend/client/hostendpoint.go
```diff
@@ -0,0 +1,52 @@
...
+// (wildcarding mising fields)
+func (h *hostEndpoints) List(hostname, name *string) ([]HostEndpoint, error) {
```
fasaxc: Is using a pointer for an optional parameter a Go idiom?  In the stdlib I've seen
- multiple functions with different names, e.g. in the regexp package, you've got Find*
- using varargs to let you specify only what you want
- using an options struct to let you specify what you want

lib/backend/client/hostendpoint.go
```diff
@@ -0,0 +1,52 @@
...
+	return nil, nil
+}
+
+// Get returns information about a particular host endpoint.
+func (h *hostEndpoints) Get(hostname, name string) (*HostEndpoint, error) {
+	return nil, nil
+}
+
+// Create creates a new host endpoint.
+func (h *hostEndpoints) Create(hostname, name string, data *HostEndpoint) error {
```
fasaxc: Aren't we embedding the hostname and name in the HostEndpoint struct?  Why do we need to pass them in too?

lib/client/client.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Load the client config from the specified file (if specified) and from environment
+// variables.  The values from both locations are merged together, with file values
+// taking precedence).
+//
+// Returns a connected client.
+func LoadClientConfig(f *string) (*CalicoClientConfig, error) {
+	var c CalicoClientConfig
+
+	// Load client config from environment variables first.
```
fasaxc, line 68: Shouldn't env vars take precedence over file vars?  File is more permanent whereas env vars are what we use to override them for one invocation.

lib/client/hostendpoint.go
```diff
@@ -0,0 +1,53 @@
+package client
+
+import (
+	. "github.com/projectcalico/libcalico/lib/api"
+)
+
+// HostEndpointInterface has methods to work with HostEndpoint resources.
+type HostEndpointInterface interface {
```
fasaxc, line 12: I don't think you should put "Interface" on the end of an interface name.  What is this, TheJavaProgrammingLanguage?

lib/common/validate.go
```diff
@@ -0,0 +1,113 @@
...
+var actionRegex *regexp.Regexp
+var protocolRegex *regexp.Regexp
+
+func init() {
+	// Initialise static data.
+	config := &validator.Config{TagName: "validate", FieldNameTag: "json"}
+	validate = validator.New(config)
+
+	var err error
+	if nameRegex, err = regexp.Compile("[a-zA-Z0-9_-]+"); err != nil { panic(err) }
```
fasaxc: I've been doing `var fooRegex = regext.MustCompile(...)`  think that saves the err check

lib/backend/client/client.go
robbrockbank: oops - that should not be in there!  blatant cut and paste error

lib/backend/client/hostendpoint.go
robbrockbank: Yup I was undecided  on this - I think an options struct would be better - especially if we want to (say) add filters etc.  I'll update.

lib/backend/client/hostendpoint.go
robbrockbank: Actually this is the backend interface (i.e. the actual etc objects).  With the definitions I currently have here it does not include the IDs - but actually maybe it should.

Let's have a chat about this.  Would be good to get your opinion - especially as it will need to tie in with the model you've been proposing for your changes.

PR #127, opened 2016-06-28
Cleanup ipam

caseydavenport:
Minimizes API surface and adds some GoDoc to the API.

### DIFF
--- a/lib/ipam/block.go
+++ b/lib/ipam/block.go
@@ -13,44 +13,44 @@ const (
 	BLOCK_SIZE = 64
 )
 
-type IPVersion struct {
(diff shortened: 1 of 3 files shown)


tomdee:
Looks good and I'm happy for you to merge. There are still a couple of bits that shouldn't be public though
- BLOCK_SIZE (maybe)
- NewBlock
- And the various Block\* errors

Or do you think that blocks can and should be exposed?

caseydavenport:
I don't think blocks need to be exposed, and we should keep from exposing them unless necessary (it may be useful to expose them some day).

I'll update.

PR #130, opened 2016-07-08
Initial commit of generic resource processing

robbrockbank:


lib/api/policy.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Register v1 structure validators to validate cross-field dependencies in any of the
+// required structures.
+func init() {
+	RegisterStructValidator(validatePolicy, Policy{})
+}
+
+func validatePolicy(v *validator.Validate, structLevel *validator.StructLevel) {
+	policy := structLevel.CurrentStruct.Interface().(Policy)
+	if policy.Metadata.Tier == DefaultTierName {
```
caseydavenport, line 65: Does this mean you can't explicitly say "put this in the default tier"?

lib/api/profile.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	. "github.com/projectcalico/libcalico/lib/api/unversioned"
+)
+
+type ProfileMetadata struct {
+	ObjectMetadata
+	Labels map[string]string `json:"labels,omitempty" validate:"omitempty,labels"`
+}
+
+type ProfileSpec struct {
```
caseydavenport, line 26: I think all these structs are our public API definition and should probably have detailed comments to describe what each struct is, and how each field behaves.

lib/api/rule.go
```diff
@@ -0,0 +1,64 @@
...
+
+	NotProtocol *Protocol `json:"!protocol,omitempty" validate:"omitempty"`
+	NotICMPType *int      `json:"!icmpType,omitempty" validate:"omitempty,gte=0,lte=255"`
+	NotICMPCode *int      `json:"!icmpCode,omitempty" validate:"omitempty,gte=0,lte=255"`
+
+	Source      EntityRule `json:"source,omitempty" validate:"omitempty"`
+	Destination EntityRule `json:"destination,omitempty" validate:"omitempty"`
+}
+
+type EntityRule struct {
```
caseydavenport, line 39: EntityRule feels like a weird name for this.  More like a RuleCriteria or something?  I don't really know what "entity rule" means.

lib/api/unversioned/types.go
```diff
@@ -0,0 +1,50 @@
...
+type TypeMetadata struct {
+	Kind       string `json:"kind" validate:"required"`
+	APIVersion string `json:"apiVersion" validate:"required"`
+}
+
+func (md TypeMetadata) GetTypeMetadata() TypeMetadata {
+	return md
+}
+
+func (md TypeMetadata) SetKind(kind string) {
```
caseydavenport, line 35: What is the reasoning for having setters / getters?  I thought in general this pattern was frowned upon in Go.

lib/client/tier.go
```diff
@@ -0,0 +1,139 @@
...
+		hl := api.NewTierList()
+		hl.Items = make([]api.Tier, 0, len(l))
+		for _, h := range l {
+			hl.Items = append(hl.Items, *h.(*api.Tier))
+		}
+		return hl, nil
+	}
+}
+
+// Get returns information about a particular tier.
```
caseydavenport, line 57: Get returns a particular tier

lib/selector/parser/ast.go
```diff
@@ -0,0 +1,274 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
```
caseydavenport, line 1: I see no need for the length of this filename to be restricted to three letters.

lib/api/policy.go
robbrockbank, line 65: Yes - but you specify a blank tier to put in the default tier

lib/api/profile.go
robbrockbank, line 26: Agreed

lib/api/unversioned/types.go
robbrockbank, line 35: You might be right - but this at least allows you to define an interface.  What's the preferred approach?  Hopefully not reflection.

lib/api/policy.go
caseydavenport, line 65: Is there a reason we wouldn't want to allow someone to say:

```
kind: Policy
metadata:
  tier: default
```

?

lib/api/unversioned/types.go
robbrockbank, line 35: Actually...  

https://golang.org/doc/effective_go.html#Getters

lib/api/rule.go
robbrockbank, line 39: I was trying to thing about what captures a source and a destination - i think entity cropped up in my inter web search and i thought that'll do for now.  We can change - but we should do soon before the API is used.

lib/api/unversioned/types.go
caseydavenport, line 35: Ah, yeah, that is what I read. I just misremembered what it said :)

This makes sense so we can use the Resource interface.  We should change `GetTypeMetadata` to `TypeMetadata` though.

lib/api/rule.go
caseydavenport, line 39: I think Entity is fine, but "EntityRule" sounds like a type of Rule, where in fact this is a component of a Rule.

Maybe RuleEntity instead? Or EntityMatch?

caseydavenport:
It's a _lot_ to review at once.

I've put some comments on here, but generally I think this looks pretty good.  I'm going to start another review of it now.

lib/api/config.go
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// Client configuration required to instantiate a Calico client interface.
+type ClientConfig struct {
```
caseydavenport, line 18: Probably want to stick some documentation on this struct as well. e.g which of these options are supported together? When are certain options mandatory?  What format should each field take?

lib/api/unversioned/types.go
```diff
@@ -0,0 +1,50 @@
...
+	md.Kind = kind
+}
+
+func (md TypeMetadata) SetAPIVersion(apiVersion string) {
+	md.APIVersion = apiVersion
+}
+
+// ---- Metadata common to all resources ----
+type ObjectMetadata struct {
+	Name string `json:"name,omitempty" validate:"omitempty,name"`
```
caseydavenport, line 45: How is this going to work for resources that don't typically have a name? e.g pools.

I suppose the name field is optional, and we'll have to ensure it isn't specified (alternatively, we make store it as metadata with the pool so it can be tagged.  Would be nice to be able to say "this is my public pool, this is my private pool", name them as such, and then be able to GET them using that name.)

PR #131, opened 2016-07-14
Go IPAM Fixes

caseydavenport:
Things left to do:
- [x] Stop logging so much to stderr
- [x] Don't require the `/calico/ipam/v2/host/<hostname>/<ipversion>/block` entries to be present.
- [ ] Fix bug with AssignIP
- [ ] Re-base on top of @robbrockbank's code and integrate with client.
  - [ ] Make etcd location configurable.


tomdee:
What's the reasoning behind using glog? And should you really be logging everything at v=2 and info?

caseydavenport:
@tomdee - I chose glog because it seems like a relatively popular logging module, and @robbrockbank  has been using it in his code I think.

I probably shouldn't be logging everything at v=2 and info.  I suppose I need to think about the different log levels, etc.  This was just quicker. :p

caseydavenport:
Kubernetes logging conventions seems sensible enough to me:

http://janetkuo.github.io/kubernetes/v1.1/docs/devel/logging.html

caseydavenport:
@tomdee - I've updated the logging levels to be more sensible.  Could you take another look?

PR #134, opened 2016-08-11
Respect endpoint name

caseydavenport:


robbrockbank:
Looks like you might need to update the **eq** method too.

Seems odd that we haven't hit issues with this before - I guess we don't read and write to the same endpoint usually?

caseydavenport:
I think it was fine before because every component was generating the endpoint ID in the same way.

Once the endpoint ID is something different it all falls apart.

caseydavenport:
@robbrockbank `__eq__` has been updated.

robbrockbank:
LGTM

PR #135, opened 2016-08-12
Update build to use jessie and add separate wheezy build container

robbrockbank:
Update calico/build to use jessie, add a new wheezy build container (for building older versions of calicoctl)


tomdee:
Lgtm

PR #136, opened 2016-08-13
Add test infrastructure to allow hostname to be overridden

robbrockbank:
We currently don't test whether setting $HOSTNAME actually works.... it didn't (see https://github.com/projectcalico/calico-containers/issues/1105).

This allows tests to override the hostname using the $HOSTNAME environment variable.

### DIFF
--- a/calico_test/tests/st/utils/docker_host.py
+++ b/calico_test/tests/st/utils/docker_host.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import logging
 import os
+import uuid
 from subprocess import CalledProcessError
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


calico_test/tests/st/utils/docker_host.py
```diff
@@ -52,6 +54,11 @@ def __init__(self, name, start_calico=True, dind=True,
         An IPv6 address value to pass to calicoctl as `--ipv6`. If left as None, no value will be passed.
         """
 
+        self._hostname = None if not override_hostname else uuid.uuid1().hex[:16]
+        """
+        Create an arbitrary hostname if we want to override.
```
ozdanborne, line 62: How bout `When True, arbitrarily sets the HOSTNAME environment variable in both the DockerHost's env at startup and before all subsequent calicoctl calls.`

calico_test/tests/st/utils/docker_host.py
```diff
@@ -178,13 +189,20 @@ def start_calico_node_with_docker(self):
...
+        # as an environment variable.
+        if self._hostname:
+            hostname_args = "-e HOSTNAME=%s" % self._hostname
+        else:
+            hostname_args = ""
+
         self.execute("docker run -d --net=host --privileged "
                      "--name=calico-node "
-                     "-e IP=%s -e ETCD_AUTHORITY=%s "
+                     "%s -e IP=%s -e ETCD_AUTHORITY=%s "
```
ozdanborne: can you pop this onto its own line, for clarity?

ozdanborne: oh. kinda wierd and annoying that we do 2 vars on the same line and the rest we don't. I guess you're just following suit. Wanna split all three of these for clarity?

ozdanborne:
Code looks rock solid. Just a few readability comments to make it easier for the next guy, then LGTM

PR #137, opened 2016-08-15
Make the override hostname public

robbrockbank:
Need to make the override hostname public so that I can check it's actually being used in the test.


ozdanborne:
LGTM

PR #138, opened 2016-08-19
Add cluster GUID

fasaxc:


calico_containers/pycalico/datastore.py
```diff
@@ -291,9 +292,25 @@ def ensure_global_config(self):
...
+        # Create our Cluster GUID (if it does not already exist).
+        self._create_cluster_guid(CLUSTER_GUID_PATH)
+
         # We are always ready.
         self.etcd_client.write(CALICO_V_PATH + "/Ready", "true")
 
+    def _create_cluster_guid(self, key):
+        """
+        Create a globally unique cluster GUID.  Write it idempotently into the
+        datastore, if an exist cluster GUID doesn not already  exist.
```
fasaxc: Typo: `if an exist cluster GUID doesn not already  exist`

calico_containers/pycalico/datastore.py
```diff
@@ -291,9 +292,25 @@ def ensure_global_config(self):
         # Disable felix status reporting, which consumes etcd write bandwidth.
         self._write_global_config(FELIX_REPORTING_INTERVAL_PATH, "0")
 
+        # Create our Cluster GUID (if it does not already exist).
+        self._create_cluster_guid(CLUSTER_GUID_PATH)
+
         # We are always ready.
         self.etcd_client.write(CALICO_V_PATH + "/Ready", "true")
 
+    def _create_cluster_guid(self, key):
```
fasaxc: It's a little subtle, but this method doesn't always create a cluster GUID; I'd be tempted to call it `_ensure_cluster_guid`

calico_containers/pycalico/datastore.py
```diff
@@ -291,9 +292,25 @@ def ensure_global_config(self):
...
+        """
+        Create a globally unique cluster GUID.  Write it idempotently into the
+        datastore, if an exist cluster GUID doesn not already  exist.
+        The prevExist=False ensures an idempotent create operation.
+        """
+        try:
+            self.etcd_client.read(key)
+        except EtcdKeyNotFound:
+            UUID = uuid.uuid4()
```
fasaxc: Nit: we generally follow PEP-8 naming/formatting conventions so UUID should be lower-case.  Note: the name uuid will clash with the module import so you could call it `guid`.

calico_containers/pycalico/datastore.py
```diff
@@ -291,9 +292,25 @@ def ensure_global_config(self):
...
+            uuid_string = UUID.get_hex()
+            self.etcd_client.write(key, uuid_string, prevExist=False)
```
fasaxc: If the CaS fails, `write()` will raise an exception (I think it'll be `EtcdCompareFailed`), which you should catch and ignore.

fasaxc: Actually, the exception might be `EtcdAlreadyExist`

calico_containers/pycalico/datastore.py
```diff
@@ -291,9 +292,25 @@ def ensure_global_config(self):
...
 
+        # Create our Cluster GUID (if it does not already exist).
+        self._create_cluster_guid(CLUSTER_GUID_PATH)
+
         # We are always ready.
         self.etcd_client.write(CALICO_V_PATH + "/Ready", "true")
 
+    def _create_cluster_guid(self, key):
+        """
+        Create a globally unique cluster GUID.  Write it idempotently into the
```
fasaxc: Again, fairly subtle, but I think you could make this comment more accurate by saying 

```
Ensures that etcd contains a GUID for the cluster.
Creates the GUID (safely with CaS) if it doesn't exist.
```

fasaxc:
General feedback:
- Since this is the first code I've reviewed for you, I've erred on the side of commenting on every detail that I noticed.  Please don't be put off by the volume of comments!
- Feel free to put up a PR earlier in the process.  That lets our CI tests run against the code and enables the github code-review tools.  We often flag in-progress changes with [do not merge] or [WiP] in the title; but in general, no-one's going to review your code until you ask for a code review anyway.
- We have good UT coverage in libcalico, please can you fix up the UTs and add some UT for the new feature.  You should make sure you know how to run the UTs locally too.
- I noticed the commit history was just a series of "dev work" commits.  Do you have a plan to clean those up?  (I'd like to have a more informative history by the time this gets merged.)
- I recommend creating a branch even though you're in your own fork.  It makes it much easier to keep track once you have more than one set of changes in flight.

calico_containers/pycalico/datastore.py
mikev: Will do...

I've spent the past two years programming in Erlang, where the convention is the opposite :-)

calico_containers/pycalico/datastore.py
mikev: fixed.

calico_containers/pycalico/datastore.py
mikev: Changed name.

calico_containers/pycalico/datastore.py
```diff
@@ -291,9 +292,31 @@ def ensure_global_config(self):
         # Disable felix status reporting, which consumes etcd write bandwidth.
         self._write_global_config(FELIX_REPORTING_INTERVAL_PATH, "0")
 
+        # Create our Cluster GUID (if it does not already exist).
+        self._ensure_cluster_guid(CLUSTER_GUID_PATH)
```
robbrockbank: Why not just use _write_global_config() here rather than implementing a new method to do the same thing.  It means you have to create a UUID regardless of whether you need to, but that shouldn't be much of an overhead for start of day.

calico_containers/pycalico/datastore.py
```diff
@@ -291,9 +292,31 @@ def ensure_global_config(self):
...
+            self.etcd_client.read(key)
+        except EtcdKeyNotFound:
+            guid = uuid.uuid4()
+            guid_string = guid.get_hex()
+            try:
+                self.etcd_client.write(key, guid_string, prevExist=False)
+            except EtcdAlreadyExist:
+                # ignore
+                pass
+            else:
```
robbrockbank: This else branch is redundant.

calico_containers/pycalico/datastore.py
fasaxc: Does `_write_global_config` bail out if the config already exists?  We don't want to update the UUID if it's already there.

Issue #140, opened 2016-08-24
Default ASN is Not in Private Range


gmcquillan:
It's using 64511, where the private range begins at 64512 [1].

Value as seen in Etcd:

```
ip-10-139-4-50 ~ # etcdctl get /calico/bgp/v1/global/as_num
64511
```

From the RFC:

```
IANA has reserved, for Private Use, a contiguous block of 1023
Autonomous System numbers from the "16-bit Autonomous System Numbers"
registry, namely 64512 - 65534 inclusive.
```

Where this default value lies: 

https://github.com/projectcalico/libcalico/blob/d9fe4aa49dd193c313b481c2e5a138601e3b8b6e/calico_containers/pycalico/datastore.py#L95

1: https://tools.ietf.org/html/rfc6996

fasaxc:
It's not quite that bad: 64511 is in the "Reserved for use in documentation and sample code" block so it's not public either:

http://www.iana.org/assignments/as-numbers/as-numbers.xhtml

That said, I agree that we shouldn't be using an AS from that block since that's not what it's for.

gmcquillan:
Cool. I half expected that I was reading the RFC wrong. Cheers!

Thanks Shaun

caseydavenport:
This should be fixed in the latest beta, and will be included in Calico v2.0.0

PR #141, opened 2016-08-24
Move Default ASN Into Private 16 Bit Range.

gmcquillan:
This should help prevent any conflicts if any of these private clusters were to
become accessible directly from the internet.

Should resolve ticket #140.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -92,7 +92,7 @@ LOG_FILE_PATH_PATH = CONFIG_PATH + "LogFilePath"
 FELIX_REPORTING_INTERVAL_PATH = CONFIG_PATH + "ReportingIntervalSecs"
 
 # The default node AS number.
-DEFAULT_AS_NUM = 64511
+DEFAULT_AS_NUM = 64512
 
 # The default node mesh configuration.
 DEFAULT_NODE_MESH = {"enabled": True}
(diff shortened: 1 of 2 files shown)


matthewdupre:
Sorry it took so long for me to look at this!

I'm just thinking about the upgrade impact (make cluster with old version, then add a node with new version).  I'm reluctant to add a manual upgrade step; but I think it might not be needed.  With me to test (or you could if you like @gmcquillan ).

gmcquillan:
Yeah, that makes sense. It's a delicate thing. Seems to me that the safest upgrade path would be to make new nodes on a new ASN, and tell calico to peer them? I believe that'd result in a full mesh. I don't know if there'd be any major downsides, though.

matthewdupre:
@gunjan5 maybe we can fix this default value in 2.0?

robbrockbank:
Once the global value has been set by the calico/node image, it won't be modified again - so starting up a new node image with a different "starting" AS number will make no difference.  The global default is actually configurable, so if we went and changed it under peoples feet it would be a bad thing :-)

I think making this change should be fine - just requires docs updates, and a change to the libcalico-go to also use the new default.   The change to libcalico-go just changes what value the global default will be set to if a user "resets" the global default AS (which they can do using `calicoctl config unset`)

robbrockbank:
@gmcquillan :  Would it be possible to rebase this as it looks like there is now a merge conflict.

I'm good to press the button once that is done.

gunjan5:
We updated the default AS number to 64512 in the `2.0` release since it was a major release. PR: https://github.com/projectcalico/libcalico-go/pull/273 
I think we should close this PR now that the major change has been added to the latest release.

Re-open this PR if anyone thinks this should also be merged into the python versions.

PR #142, opened 2016-09-01
Fix calico support for the etcd3 server.  Use specific commit of python-etcd client.

mikev:


mikev:
Per Tom, we are only modifying build-requirements-frozen.txt  build-requirements-frozen is normally auto-generated, but we are overriding the generated value until the python-etcd patch becomes available in a numbered version release.  In any case, the generated value is only generated if we trigger it.

setup.py
```diff
@@ -57,6 +57,6 @@
     # your project is installed. For an analysis of "install_requires" vs pip's
     # requirements files see:
     # https://packaging.python.org/en/latest/requirements.html
-    install_requires=['netaddr', 'python-etcd>=0.4.2', 'subprocess32'],
+    install_requires=['netaddr', 'python-etcd>=0.4.3', 'subprocess32'],
```
tomdee, line 60: This is going to be a problem for calico-node

PR #143, opened 2016-09-01
add python egg info to git URL for python-etcd

mikev:


mikev:
Missed some crucial bits.  The previous commit is harmless however.

PR #144, opened 2016-09-02
Add warning on libcalico requirements file.

ozdanborne:
Add information stated in #142 to requirements file so it's better tracked.


mikev:
I double-checked that comments are allowed - https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format

lgtm.

Issue #148, opened 2016-09-27
Invalid mac address returned if `ip` prints to stderr


cutwater:
Function `get_ns_veth_mac` [netns.py#L245](https://github.com/projectcalico/libcalico/blob/master/calico_containers/pycalico/netns.py#L245) returns invalid result if `ip netns ...` command printed message in stderr.

`ip netns` command can print `RTNETLINK answers: Invalid argument` to stderr and still finish with exitcode `0`, which causes next endpoint written to etcd:

`{"name": "cali76e50b0684a", "labels": {}, "state": "active", "ipv6_nets": [], "mac": "RTNETLINK answers: Invalid argument\n2a:1b:78:a8:63:a4", "ipv4_nets": ["10.233.67.164/32"], "profile_ids": ["calico-k8s-network"]}`

The root cause of this is that function [NamedNamespace.check_output](https://github.com/projectcalico/libcalico/blob/master/calico_containers/pycalico/netns.py#L345) redirects **stderr** to **stdout**.

Solution is to read output only from **stdout**.

matthewdupre:
Assigning this a low priority on the assumption that the original cause (of ip netns erroring) can be worked around: I can adjust if it can't.  @cutwater did you find out why the command errored?

cutwater:
@matthewdupre We're still investigating the root cause. This issue occurs when there is something wrong with linux namespaces (usually after OpenStack neutron creates `dhcp-` namespace). Still `ip netns` works correctly, returns the result as expected and exitcode is `0`, except it has such side-effect like writing message to _stderr_.

The problems begin when `NamedNamespace` reads stderr, considering it as a result of program execution. In my opinion this is incorrect behaviour, which affects data that is written to etcd.

nelljerram:
@djosborne You added the `stderr=STDOUT` as part of:

```
65e86e67980e6efe7461b5ea6427067e2f45c332
Author:     Dan Osborne <daniel.osborne@metaswitch.com>
AuthorDate: Tue Oct 13 16:27:16 2015 -0700
Commit:     Dan Osborne <daniel.osborne@metaswitch.com>
CommitDate: Tue Oct 13 16:27:16 2015 -0700

Removed use of check_call in netns and improved logging

Replaced with check_output, which, when it throws an exception, includes a .output field which can be used to better log encountered errors
```

Can you remember why?

ozdanborne:
[...]

PR #149, opened 2016-09-27
Remove redirect stderr to stdout in NamedNamespace

cutwater:
Fixes issue #148

### DIFF
--- a/calico_containers/pycalico/netns.py
+++ b/calico_containers/pycalico/netns.py
@@ -21,7 +21,7 @@ import uuid
 import re
 from copy import copy
 
-from subprocess32 import check_output, check_call, CalledProcessError, STDOUT
+from subprocess32 import check_output, check_call, CalledProcessError
 from netaddr import IPAddress
 
 from util import get_ipv6_link_local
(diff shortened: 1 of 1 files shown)


calico_containers/pycalico/netns.py
```diff
@@ -342,7 +342,7 @@ def check_output(self, command):
         """
         command = self._get_nets_command(command)
         _log.debug("Run command: %s", command)
```
nelljerram, line 344: Please let's add a comment here:

```
# Note: we must not also capture the stderr here (as with
# stderr=STDOUT), because an otherwise successful 'ip netns exec <ns>
# <command> ...' call will write error messages to stderr if there are
# any bogus namespaces on the system.  See
# https://github.com/projectcalico/libcalico/issues/148 for detailed
# discussion of this.
```

nelljerram (changes requested):
Please add the comment that I've suggested, then this will be good to go.

calico_containers/pycalico/netns.py
cutwater: Done

nelljerram (approved):
Thanks!

PR #151, opened 2016-10-03
Fix multiple prefix support in libcalico

cutwater:
Function `ensure_global_config` now appends interface prefix for calico
if it is not present.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -261,7 +261,7 @@ class DatastoreClient(object):
         :return: None.
         """
         # Configure Felix config.
-        self._write_global_config(CONFIG_IF_PREF_PATH, IF_PREFIX)
+        self._ensure_interface_prefix()
 
         # Configure IPAM directory structures (to ensure confd is able to
         # watch appropriate directory trees).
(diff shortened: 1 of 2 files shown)


calico_containers/pycalico/datastore.py
```diff
@@ -298,6 +298,18 @@ def ensure_global_config(self):
         # We are always ready.
         self.etcd_client.write(CALICO_V_PATH + "/Ready", "true")
 
+    def _ensure_interface_prefix(self):
+        try:
+            if_prefix = self.etcd_client.read(CONFIG_IF_PREF_PATH).value
```
nelljerram, line 303: Move `prefixes = ...` line here.

calico_containers/pycalico/datastore.py
```diff
@@ -298,6 +298,18 @@ def ensure_global_config(self):
...
+        except EtcdKeyNotFound:
+            self.etcd_client.write(CONFIG_IF_PREF_PATH, IF_PREFIX)
+            return
```
nelljerram: Delete previous two lines, and instead put `prefixes = []`

nelljerram (commented):
This basically LGTM - I've just suggested a couple of slight simplifications to the code.  WDYT?

calico_containers/pycalico/datastore.py
cutwater: Done

calico_containers/pycalico/datastore.py
cutwater: Done.

nelljerram (approved):
Thanks!

PR #154, opened 2016-10-05
[WIP] Changes to enable ST tests of golang calicoctl

lwr20:


Dockerfile.calico_test
```diff
@@ -62,6 +62,10 @@ RUN pip install -r /tmp/pycalico/calico_test/requirements.txt
 # Add the testing framework
 ADD calico_test/tests tests
 
+# Install etcdctl
+RUN wget https://github.com/coreos/etcd/releases/download/v2.3.3/etcd-v2.3.3-linux-amd64.tar.gz && tar -xzf etcd-v2.3.3-linux-amd64.tar.gz
+RUN cd etcd-v2.3.3-linux-amd64 && mv etcdctl /usr/local/bin/
```
tomdee, line 67: Better to just create a symlink here, or do the move on the same image layer as above.

tomdee (commented):
There's a lot of noise from the formatting changes - would it be possible to put those in a separate PR?

PR #155, opened 2016-10-07
Fix missing import

cutwater:
EtcdAlreadyExist name is used but not imported, which causes
NameError.


cutwater:
Closes #156

caseydavenport:
@cutwater Thanks!  LGTM.

PR #161, opened 2016-10-12
Don't overwrite DefaultEndpointToHostAction when running calicoctl node if it was already set

matthewdupre:
Fixes https://github.com/projectcalico/calico-containers/issues/1190

I've left the overwrite other config code in place for now: it shouldn't cause problems because they can be specified when running `calicoctl node`.


tomdee:
LGTM

PR #162, opened 2016-10-14
Code tidy up

lwr20:


tomdee:
I've taken a quick look, and on the whole I support this change. However, I'm not sure about all the line splitting. What line width are you working to?

lwr20:
My understanding was that for python, we are working to PEP8 - so 80 columns.

PR #165, opened 2016-10-27
Update frozen requirements.

ozdanborne:
Notably this uses nonforked python-etcd.

Generated just by running `make update-frozen`


tomdee:
Erm... quite a lot going on here. Have you thought about all these changes and you'll vouch for them all?

ozdanborne:
@tomdee I misunderstood the purpose of having requirements and frozen-requirements. Updated to just update etcd in this PR.

I don't actually see the purpose of having requirements.txt anymore. Doesn't seem like anything uses or builds from it.

caseydavenport:
@djosborne I'm assuming you've verified v0.4.2 includes the commit that was pinned before?

ozdanborne:
I assumed since 0.4.2 was released a month after the necessary commits, it would.

But shouldn't have assumed. The necessary commits weren't included.

PR #166, opened 2016-11-07
Fix KeyError: 'affinity'

CaoShuFeng:
When etcd doesn't store key "affinity", command "calicoctl node remove"
fails with KeyError. This change uses dict.get("affinity") instead
dict["affinity"] to avoid keyrror.

Fixes projectcalico/calico-containers#1279

### DIFF
--- a/calico_containers/pycalico/block.py
+++ b/calico_containers/pycalico/block.py
@@ -147,7 +147,7 @@ class AllocationBlock(object):
         # Parse out the host.  For now, it's in the form host:<host id>.  An
         # empty host ID is converted to None to indicate the block has no
         # specific host affinity.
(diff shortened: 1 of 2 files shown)


caseydavenport:
This is related to this PR: https://github.com/projectcalico/libcalico-go/issues/226

Basically the Go code very briefly wrote affinity to etcd with the wrong format, and that caused the Python code to barf on it.

Not sure if we need to handle the `hostAffinity: host` case at all in the Python code, since it's the Go IPAM driver that is writing out the wrong information, and people (probably?) aren't running the Python and the Go IPAM drivers together.

Worst comes to worst, if someone uses both together, then the Python code might not respect block affinity in some cases.

caseydavenport:
LGTM.

PR #167, opened 2016-11-07
Parallelise connectivity checks for speed

lwr20:
STs which involve a lot of connectivity checking are taking too long.  In particular, tests which we expect to fail take 1 second to timeout, because that's the minimum timeout.  We should do connectivity checks in parallel to save time.


matthewdupre (approved):
LGTM

PR #168, opened 2016-11-11
Improve ST diags

lwr20:
Output more diags for failures and make the output easier to read with banners for key moments in the test in order to address https://github.com/projectcalico/calico-containers/issues/1270

Also move the code which wiped etcd after every test to run once per testClass instead to allow groups of tests to setup shared test setup.

### DIFF
--- a/calico_test/tests/st/utils/utils.py
+++ b/calico_test/tests/st/utils/utils.py
@@ -133,13 +133,30 @@ class TestWithHooks(unittest.TestCase):
 
     @classmethod
     def get_diags(cls):
+        logfiles = [
+            "/var/log/calico/bird/current",
+            "/var/log/calico/bird6/current",
+            "/var/log/calico/confd/current",
+            "/var/log/calico/felix/current",
+            "/var/log/calico/libnetwork/current",
+        ]
         if cls.hosts:
             for host in cls.hosts:
                 if host.dind:
-                    logger.debug("Docker logs from %s (%s):\r\n%s",
-                                 host.name,
-                                 host.ip,
-                                 host.execute("docker logs calico-node"))
+                    try:
+                        logger.debug("Docker logs from %s (%s):\r\n%s",
+                                     host.name,
+                                     host.ip,
+                                     host.execute("docker logs calico-node"))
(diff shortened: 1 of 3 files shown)


calico_test/tests/st/utils/utils.py
```diff
@@ -33,6 +38,125 @@
...
+    @classmethod
+    def on_failure(cls, test_id=None):
+        """
+        The test raised an exception.
+        """
+        cls.get_diags()
+        logger.info("Test failed.")
+
+    @classmethod
+    def get_diags(cls):
```
robbrockbank, line 135: This code does not belong here - it invokes processing that is specific to the derived class (i.e. TestBase).  It should be in TestBase, perhaps this class just needs a no-op method.

calico_test/tests/st/utils/utils.py
```diff
@@ -33,6 +38,125 @@
...
+        logfiles = [
+            "/var/log/calico/bird/current",
+            "/var/log/calico/bird6/current",
+            "/var/log/calico/confd/current",
+            "/var/log/calico/felix/current",
+            "/var/log/calico/libnetwork/current",
+        ]
+        if cls.hosts:
+            for host in cls.hosts:
```
robbrockbank: Doesn't this get invoked *after* the test has failed ... in which case the DockerHosts have exited and the containers won't be around anymore?

robbrockbank:
Lance, back with you, a couple of points in the review.

calico_test/tests/st/utils/utils.py
lwr20: It does.  You'll see in the matching calico-containers PR that I've moved from a `with Dockerhost..` to using SetUp to create and TearDown to delete containers.

calico_test/tests/st/utils/utils.py
lwr20: Moved.  And left a no-op method as suggested.

lwr20:
OK - changes made.  @robbrockbank Can you re-review please?

lwr20:
And rebased onto latest master

PR #174, opened 2016-12-06
Move per-host config out of global config

robbrockbank:
IPAM affinity block directories were not being initialized using the correct host name due to HOSTNAME->NODENAME environment change in startup.py in calico/node.

Problem is because we put some per-host init in the global init rather than in the host creation.  This PR just moves it out.


caseydavenport:
LGTM

PR #177, opened 2017-01-11
Retry workload create and network create

lwr20:


matthewdupre (approved):
LGTM - let's see if it helps.

PR #181, opened 2017-01-19
Modify default log format for container tests

lwr20:
Format has been taken from the Felix log formatter (@fasaxc recommended it). 

Example output with the new formatter:
```
2017-01-19 12:49:18,801 [INFO][26158/140043457387184] tests.st.utils.utils 73:   # 64 bytes from 192.168.141.192: seq=0 ttl=64 time=0.145 ms
2017-01-19 12:49:18,802 [INFO][26158/140043457387184] tests.st.utils.utils 73:   # 
2017-01-19 12:49:18,803 [INFO][26158/140043457387184] tests.st.utils.utils 73:   # --- 192.168.141.192 ping statistics ---
2017-01-19 12:49:18,803 [INFO][26158/140043457387184] tests.st.utils.utils 73:   # 1 packets transmitted, 1 packets received, 0% packet loss
2017-01-19 12:49:18,804 [INFO][26158/140043457387184] tests.st.utils.utils 73:   # round-trip min/avg/max = 0.145/0.14
```

Timestamp should allow us to do things like profiling tests (seeing where all the time is going).

### DIFF
--- a/calico_test/tests/st/test_base.py
+++ b/calico_test/tests/st/test_base.py
@@ -27,8 +27,9 @@ from tests.st.utils.utils import (get_ip, ETCD_SCHEME, ETCD_CA, ETCD_CERT,
 
 HOST_IPV6 = get_ip(v6=True)
 HOST_IPV4 = get_ip()
-FORMAT_STRING = '%(asctime)s [%(levelname)s][%(process)s/%(thread)x] %(name)s %(lineno)d: %(message)s'
-logging.basicConfig(level=logging.DEBUG, format=FORMAT_STRING)
+FORMAT_STRING = '# %(asctime)s [%(levelname)s][%(process)s/%(thread)x] ' \
+                '%(name)s %(lineno)d: %(message)s'
+logging.basicConfig(level=logging.DEBUG, format=FORMAT_STRING, datefmt='%H:%M:%S')
 logger = logging.getLogger(__name__)
 
 # Disable spammy logging from the sh module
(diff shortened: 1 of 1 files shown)


fasaxc:
I consider the timestamp, severity and file/line no must-haves.  Rest is useful from time to time.

tomdee:
Some of the test output is designed to be copy pastable to repro tests. E.g. some lines start with a hash and some don't. Will this break that?

lwr20:
@tomdee Ah - now that's exactly the sort of thing I was hoping to find out. Where does that hash get put on?

tomdee:
You should be able to find it by looking at the logging code

lwr20:
That's not very helpful - there's "logging code" throughout the tests.  Ah - I think you're referring to `log_and_run()` in `utils.py`.  

However, looking at the other various instances of logging around the place, I'm fairly sure that the requirement that the output be cut-and-pastable wasn't understood by authors and reviewers for quite a few PRs now - its clearly been broken for a long time.

Before I spend time fixing them all up, is this still a requirement?

fasaxc:
@lwr20 Can you just add a `#` to the start of the log format, then make sure that `log_and_run` puts a `\n\n` before the commands it prints?

lwr20:
Ah - good call.   I'll also need to fix up all the places we've broken it in the past, but yeah - maybe that's doable.

lwr20:
Hmm.  Not sure that's working quite right.   Commands are getting timestamped.  Let me have a further think...

tomdee:
It's not a requirement. I was just letting you know that it was once a thing.

lwr20:
Moved to https://github.com/projectcalico/calicoctl/pull/1472

PR #185, opened 2017-04-04
Rev requests and urllib3.

### LINKED ISSUE #69: Policy controller doesn't support mixed IP SANs/DNS SANs
This often trips people up, especially if they're following a perfectly reasonable guide to get k8s certs set up then they try to add Calico.

I believe IP SAN support has now landed in urllib3 so revving that library might get us that support "for free".

fasaxc:
- Rev urllib3 to latest (1.20.0) to pick up security fixes.
- Rev requests to 2.13.0 pick up newer vendored version of urllib3; includes the security fixes above as well as support for IP SANs. Should fix https://github.com/projectcalico/k8s-policy/issues/69.

This passed the policy agent STs but I don't think they test TLS.  Since the fix targets TLS, feels like we should test that before merging.  Will follow up on slack to figure out the best way to do that.

### DIFF
--- a/build-requirements-frozen.txt
+++ b/build-requirements-frozen.txt
@@ -9,9 +9,9 @@ PyInstaller==3.1.1
 git+https://github.com/jplana/python-etcd.git@0d0145f5e835aa032c97a0a5e09c4c68b7a03f66
 prometheus_client==0.0.13
 PyYAML==3.11
-requests==2.9.1
+requests==2.13.0
 six==1.10.0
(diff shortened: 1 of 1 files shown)


fasaxc:
Ah, Casey's on holiday, moving review request to Tom.

tomdee (approved):
LGTM - but still need to do a manual test

fasaxc:
Customer has confirmed this fixes the issue.

